PR #6, opened 2025-12-06
Fix cache key to include v2 fields (container, env, mounts)

yairfalse:
## Summary
- Fixed cache key generation to include v2 SDK fields (container, task.env, mounts)
- Previously, changing container image (e.g., golang:1.21 → golang:1.22) would NOT invalidate cache
- Added `hash_task_env/1` and `hash_mounts/1` helper functions for deterministic hashing
- Refactored tests to use `make_task/3` helper for cleaner test code

## Test plan
- [x] All 21 cache tests pass
- [x] All 23 Elixir tests pass
- [x] All 37 Go SDK tests pass
- [x] New tests verify cache invalidation when:
  - Container image changes
  - Task env vars change
  - Mount configuration changes

### DIFF
--- a/core/lib/sykli/cache.ex
+++ b/core/lib/sykli/cache.ex
@@ -282,7 +282,7 @@ defmodule Sykli.Cache do
   defp hash_mounts([]), do: ""
   defp hash_mounts(mounts) do
     mounts
-    |> Enum.map(fn m -> "#{m.resource}:#{m.path}:#{m.type}" end)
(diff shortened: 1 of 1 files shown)


yairfalse:
Root cause: When a port receives {:exit_status, status}, Erlang automatically closes it. The after block was then calling Port.close(port) on an already-closed port, raising ArgumentError.

  Fix: Added safe_port_close/1 helper that catches ArgumentError:

  defp safe_port_close(port) do
    try do
      Port.close(port)
    rescue
      ArgumentError -> :ok
    end
  end

PR #114, opened 2026-04-03
feat(cli): shared JSON envelope for all --json output

yairfalse:
## Summary
- All 9 `--json` commands now return `{"ok", "version", "data", "error"}` envelope
- Agents get one shape to parse, regardless of command
- Error responses carry `code`, `message`, `hints` from `Sykli.Error`
- New `Sykli.CLI.JsonResponse` module with `ok/1` and `error/1`

## Contract

```json
{"ok": true,  "version": "1", "data": { ... }, "error": null}
{"ok": false, "version": "1", "data": null,    "error": {"code": "...", "message": "...", "hints": [...]}}
```

## Test plan
- [x] `mix test` — 1075 tests, 0 failures
- [x] `test/blackbox/run.sh` — 89/89 pass (5 new envelope tests added)
- [x] `mix escript.build` — binary builds clean

### DIFF
--- a/core/lib/sykli/cli.ex
+++ b/core/lib/sykli/cli.ex
@@ -773,7 +773,13 @@ defmodule Sykli.CLI do
     case Sykli.Validate.validate(path) do
       {:ok, result} ->
         if json_output do
-          IO.puts(JsonResponse.ok(Sykli.Validate.to_map(result)))
+          data = Sykli.Validate.to_map(result)
+
+          if result.valid do
+            IO.puts(JsonResponse.ok(data))
+          else
+            IO.puts(JsonResponse.error_with_data(data))
+          end
         else
           output_validation_result(result)
         end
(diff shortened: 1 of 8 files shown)


core/lib/sykli/cli.ex
```diff
@@ -773,20 +773,24 @@ defmodule Sykli.CLI do
...
+          data = Sykli.Validate.to_map(result)
+
+          if result.valid do
+            IO.puts(JsonResponse.ok(data))
+          else
+            IO.puts(JsonResponse.error_with_data(data))
+          end
         else
           output_validation_result(result)
         end
```
yairfalse, line 786: All 4 comments addressed in 70b02e5.

**ok vs exit code (cli.ex:786):** Fixed. `validate --json` now returns `ok: false` when `result.valid` is false. Added `JsonResponse.error_with_data/1` for this case — the envelope signals failure but still carries the validation data so agents can inspect errors.

**Plan version assertions (dataset.json:203, :443):** Added `"version": "1"` to POS-023 and SYS-011.

**Fix test variable naming (fix_test.exs:184):** Renamed `json` → `result` to match the `to_map` rename.

PR #115, opened 2026-04-04
feat(cli): sykli run --json structured output

yairfalse:
## Summary
- `sykli run --json` emits clean JSON on stdout — no ANSI, no progress lines
- `ok: true` with `status: "passed"/"failed"` for completed runs
- `ok: false` with structured error code for infra failures
- Stdout suppressed during execution via group_leader redirect

## Contract
```json
{"ok": true, "version": "1", "data": {"status": "passed", "duration_ms": 420, "tasks": [{"name": "test", "status": "passed", "duration_ms": 180, "cached": false}], "occurrence_path": ".sykli/occurrence.json"}, "error": null}
```

## Test plan
- [x] `mix test` — 1076 tests, 0 failures
- [x] `test/blackbox/run.sh` — 92/92 pass (3 new run --json cases)
- [x] Manual: pass, fail, and infra-error all produce clean JSON

### DIFF
--- a/core/lib/sykli/cli.ex
+++ b/core/lib/sykli/cli.ex
@@ -137,26 +137,32 @@ defmodule Sykli.CLI do
   defp run_sykli(args) do
     {path, opts} = parse_run_args(args)
     start_time = System.monotonic_time(:millisecond)
-
-    # Build run options
-    run_opts = []
-
-    # Handle K8s target - prepare git context first
     json_output = Keyword.get(opts, :json, false)
 
-    case prepare_target_context(path, opts) do
-      {:ok, target_opts} ->
-        run_opts = target_opts ++ run_opts
-        do_run_sykli(path, opts, run_opts, start_time, json_output)
+    # Suppress stdout before any work so target setup lines don't leak
+    original_gl = if json_output, do: suppress_stdout(), else: nil
 
-      {:error, reason} ->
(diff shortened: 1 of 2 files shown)


core/lib/sykli/cli.ex
```diff
@@ -136,19 +137,32 @@ defmodule Sykli.CLI do
...
-        do_run_sykli(path, opts, run_opts, start_time)
+    try do
+      run_opts = []
 
-      {:error, reason} ->
-        display_error(reason)
-        halt(1)
+      case prepare_target_context(path, opts) do
+        {:ok, target_opts} ->
+          run_opts = target_opts ++ run_opts
```
yairfalse, line 150: All 3 comments addressed in 9fd1751.

**Suppression before target setup (cli.ex:150):** Moved `suppress_stdout()` into `run_sykli/1` before `prepare_target_context`, so the K8s "Target: k8s" line and any other setup output is captured.

**Discard IO process (cli.ex:2797):** Replaced `StringIO` with a spawned IO server process that discards all writes with zero memory. Process auto-exits when parent goes down via `Process.monitor`.

**try/after guard (cli.ex:229):** Wrapped the entire run in `try/after` that calls `restore_stdout(original_gl)`, guaranteeing restoration on raise/throw/exit.

PR #134, opened 2026-05-01
chore: complete v0.6.0 release + close docs gaps

yairfalse:
Summary: closes the v0.6.0 release/docs drift so the release, docs, ADR state, and installer path can converge on 0.6.0.\n\nScope:\n- Reframes CLAUDE.md around ADR-022's execution graph compiler positioning.\n- Updates README hero/version/install table/CLI command list and marks the hero review task as roadmap.\n- Accepts ADR-022 and marks ADR-020 positioning as superseded while preserving visual direction.\n- Regenerates sdk/typescript/package-lock.json root package metadata to 0.6.0.\n- Fixes release workflow blockers found on the first v0.6.0 tag run: macOS x86 runner label, pinned macOS Zig 0.15.2, release notes from docs/releases/v0.6.0.md, and Latest marking.\n- Makes Credo available at compile time for prod release builds while keeping it runtime:false, fixing the failed Burrito build.\n\nVerification:\n- mix format\n- mix test: 1433 tests, 0 failures, 1 skipped, 13 excluded\n- mix credo: no issues\n- mix escript.build\n- ./core/sykli --help\n- ./core/sykli --version -> sykli 0.6.0\n- test/blackbox/run.sh: 137 passed, 11 expected-red, 0 failed\n- Prod release smoke: local MIX_ENV=prod BURRITO_TARGET=linux_x86_64 mix release now compiles through sykli; local wrap stops only because local Zig is 0.16.0 and Burrito requires 0.15.2. Workflow now pins 0.15.2 on Linux and macOS.\n\nRelease status:\n- CLAUDE_CODE_OAUTH_TOKEN secret: set in repo Actions secrets.\n- v0.6.0 tag currently exists from the first attempt, but the workflow failed before release creation. After this PR merges, retag v0.6.0 at the merged HEAD and push the corrected tag so release.yml creates the GitHub Release from docs/releases/v0.6.0.md.\n- gh release view v0.6.0 currently reports release not found; v0.5.2 remains Latest until the corrected tag run succeeds.\n\nNotes:\n- The broad acceptance grep for 0.5.1 still matches a transitive npm package, local-pkg-0.5.1, inside package-lock.json. The stale sykli package-lock root version is fixed to 0.6.0; I did not alter unrelated transitive dependency metadata.

PR #135, opened 2026-05-01
test: smoke Claude code review workflow

yairfalse:
Tiny no-op PR to verify claude-code-review.yml runs end-to-end with the configured slash-command prompt. This PR is for workflow verification only and should be closed after the check result is captured.


yairfalse:
Workflow smoke result captured in #134. Closing this verification-only PR.

PR #140, opened 2026-05-01
feat(graph): add review nodes

yairfalse:
## Summary
Adds the first internal model for SYKLI Reviews as non-deterministic graph nodes. This only models review work in the graph; it does not call any provider or implement review execution.

## Scope
- Adds review node metadata to the core graph parser: kind, primitive, agent, context, outputs, dependencies, deterministic flag.
- Lets validation accept commandless review nodes while preserving normal task command validation.
- Exposes review metadata in plan/explain output, including deterministic=false by default.
- Adds Go SDK support via s.Review(...).Primitive(...).Agent(...).Diff(...).Context(...).After(...).Outputs(...).
- Updates README docs to mark SYKLI Reviews experimental and distinguish task nodes from review nodes.

## Verification
- cd core && mix format
- cd core && mix test
- cd core && mix credo
- GOCACHE=/tmp/sykli-go-build-cache go test ./... in sdk/go
- git diff --check
- test/blackbox/run.sh: 137 passed, 11 expected-red, 0 failed

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,8 @@ The engine runs on the BEAM VM. Same code on your laptop, in Docker, on Kubernet
 
 SYKLI Reviews are experimental. A review node represents a structured review step in the execution graph; it does not yet run Codex, Claude, or any other provider directly. It models the review step so future runners can execute agents in a controlled, inspectable way.
 
+The builder API is currently available in the Go SDK only. Rust, TypeScript, Elixir, and Python SDK builders still need parity work; until then, review nodes are an experimental Go-first graph feature.
+
 Agentic workflows need primitives, not prompts. Asking an LLM to "review this PR" is too underspecified to be repeatable. Defining a review node with constrained inputs, expected outputs, and explicit rules is.
 
 ```go
--- a/README.md
+++ b/README.md
@@ -21,11 +21,13 @@ func main() {
         After("test").
         Outputs("app")
 
-    // roadmap: ships in 0.6.x — see Roadmap section
-    s.Task("review").
-        Run("sykli review --primitive api-breakage --diff main...HEAD").
+    s.Review("review:api-breakage").
+        Primitive("api-breakage").
+        Agent("local").
+        Diff("main...HEAD").
+        Context("README.md", "docs/architecture.md").
         After("test").
-        Outputs("review.json")
+        Outputs("reviews/api-breakage.json")
 
     s.Emit()
 }
@@ -36,12 +38,12 @@ sykli · pipeline.go                                local · 0.6.0
 
   ●  test     go test ./...                        108ms
   ●  build    go build -o app                      612ms
-  ●  review   sykli review --primitive ...         842ms
+  ○  review:api-breakage   api-breakage            planned
 
-  ─  3 passed                                      1.2s
+  ─  2 passed · 1 review planned                   720ms
 ```
 
-The `review` task is just another node in the graph. It can be executed locally, in GitHub Actions, in another CI provider, or by an agent-aware runtime. The graph doesn't care.
+The `review:api-breakage` node is not a shell task. It is a structured review node in the graph: primitive, agent identifier, diff input, context files, dependencies, outputs, and `deterministic: false`.
 
 ---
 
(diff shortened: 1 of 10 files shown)


yairfalse:
## Code Review

### Overview

Adds **review nodes** as a new graph node kind alongside tasks and gates. A review node carries metadata only (no executor): `kind: "review"`, `primitive`, `agent`, `inputs`, `context`, `outputs`, `depends_on`, `deterministic` (default `false`). Review nodes are exempt from the "must have a command" validation (matching the existing gate exemption), are never cacheable, and surface in `plan` and `explain` output.

PR description is honest: *"only models review work in the graph; does not call any provider or implement review execution."*

**Diff:** 10 files, +539 / −111. CI green across all 9 checks.

### Strengths

- **Backward compatible.** `kind` defaults to `:task` when absent in JSON, so existing pipelines parse unchanged.
- **Idiomatic accessor pattern** in `Sykli.Graph.Task`: `kind/1`, `review?/1`, `primitive/1`, `context/1`, `deterministic?/1` follow the existing `name/1`, `command/1`, `inputs/1` shape with `?`-suffixed predicates.
- **`cacheable?(%Task{kind: :review})` short-circuits to `false`** — the right call; reviews are non-deterministic by definition.
- **`validate.ex` change is one line** — minimal blast radius. The exemption mirrors the gate exemption directly above it.
- **Cycle detection in Go SDK** correctly extended: `detectCycle()` and `nodeNames()` include reviews.
- **`hasNode/1` deduplication helper** in Go SDK eliminates the previous O(n) scan duplicated across `Task`/`Gate`/`Review` constructors. Net cleanup.

### Issues

#### 🔴 SDK parity — only Go is updated

**Direct violation of CLAUDE.md:** *"When changing task schema fields, update all five SDKs and their conformance test cases."*

This PR adds `kind`, `primitive`, `agent`, `context`, `deterministic` to the schema and ships an SDK builder for them — but only in Go. Rust, TypeScript, Elixir, and Python SDKs have no `Review()` primitive. A Python user can't write `s.review("api-breakage").primitive(...)` even though the engine accepts the JSON shape. The README presents the feature as universal.

The PR description acknowledges Go-only, but the project rule has no "Go-first is OK" carve-out. **Either** ship all five SDKs in this PR **or** open follow-up tracking issues for the other four and add an explicit "experimental, Go SDK only" callout to the README. Currently the README change ("SYKLI Reviews experimental") is the closest thing, but there's no tracking for the parity gap.

#### 🟡 Brittle test — `case ... do {:error, _} -> :ok end` hides regressions

`core/test/sykli/plan_test.exs:152-192`:

```elixir
case Plan.generate(graph, from: "HEAD", path: ".") do
  {:ok, plan} -> ... assertions ...
  {:error, _} -> :ok    # ← any error silently passes the test
end
```

This pattern is used elsewhere in the same file because `Plan.generate` depends on git state. But for **this** test — which is testing schema parsing, not git state — the fall-through hides regressions. If a future change breaks `Plan.generate` for review nodes specifically, this test passes via `:ok` and the bug ships.

**Fix:** either use a `from:` value that doesn't require git, or assert the shape with `assert {:ok, plan} = Plan.generate(...)`.

#### 🟡 No conformance test cases

`tests/conformance/` is unchanged. Adding a schema element without a fixture means future SDK implementations (Rust/TS/Elixir/Python when they catch up) have nothing to validate against. Two cases would suffice: `review-minimal` (kind + primitive + agent) and `review-with-deps` (review with `after` and `context`).

#### 🟡 No black-box test cases

`test/blackbox/dataset.json` unchanged. New schema element justifies:
- POS: review node parses, validates, and shows up in `plan` output
- NEG: review missing `primitive` should fail validation (or pass — depending on intent; this PR doesn't decide)

The PR's own description shows blackbox passes 137/0/11, which means the existing suite still works — but nothing new exercises reviews.

#### 🟡 Validation is permissive on review-specific required fields

`validate.ex:232` exempts review nodes from the command check, but does **not** require `primitive` or `agent`. A review with `{"kind": "review", "name": "x"}` and nothing else passes `validate`. The future executor will have no idea what to run.

This may be intentional ("we'll enforce when we wire execution"), but the schema's contract is unclear. Either add a `:missing_review_primitive` validation, or document in `parse_review/2` that nil primitive/agent is "execution will fail" territory.

#### 🟢 Minor — `parse_kind(:review)` is dead code

```elixir
defp parse_kind("review"), do: :review
defp parse_kind(:review), do: :review     # ← never reached; JSON only produces strings
defp parse_kind(_), do: :task
```

Jason returns string keys/values; the atom clause is unreachable from the parse path. Defensive but unused.

#### 🟢 Minor — spurious smart-quote in Go docstring

`sdk/go/sykli.go:1018`: `// - tag != '' - run only when a tag is present` was changed to `tag != ”` (Unicode right double quote). Looks like an editor auto-correct slipped in. Cosmetic but should be reverted to plain ASCII.

#### 🟢 Minor — `Outputs any` in jsonTask

The Go SDK's `jsonTask.Outputs` field changed from `map[string]string` to `any` to accommodate review nodes' `[]string`. Functionally fine because of `omitempty`, but loses type safety at the marshal site. A cleaner alternative is two separate JSON struct types (`jsonTask` and `jsonReview`) — but that's a bigger refactor and the current shape works.

### Test coverage

| Test | Coverage |
|---|---|
| `core_test.exs` parse review with metadata | ✅ comprehensive |
| `validate_test.exs` exempts review from command check | ✅ |
| `plan_test.exs` review surfaces with metadata | ⚠️ brittle (see 🟡 above) |
| `sdk/go/sykli_test.go` review emits expected JSON | ✅ thorough |
| Conformance cases | ❌ none |
| Black-box cases | ❌ none |
| Cycle detection through reviews | ❌ untested in either SDK or engine |
| Review-with-missing-primitive validation | ❌ untested |

### Performance

Negligible. One added field to the Task struct, one extra branch in `parse_task`. ✓

### Security

Reviews are inert metadata in this PR — no execution path introduced. The `context` field accepts arbitrary string paths; when execution lands, it needs the same `String.starts_with?(path, base <> "/")` containment rule the rest of the codebase uses for path inputs. Not a blocker for this PR, but worth tracking.

### Recommendation

**Approve with two blocking changes:**

1. **🔴 SDK parity** — add a clear callout in README and/or open tracking issues for Rust/TS/Elixir/Python parity. The project rule is explicit; partial-SDK shipping requires explicit acknowledgment.
2. **🟡 Fix the brittle plan test** — replace `case do ... {:error, _} -> :ok end` with an assertion that fails on regression.

Items 🟡 (conformance, black-box, validation strictness) and 🟢 (dead atom clause, smart-quote, `Outputs any`) are reasonable follow-ups.

The core schema change is well-shaped, additive, and tested. The Codex tell is the brittle test pattern — a clean reviewer would have caught that.

PR #141, opened 2026-05-01
fix(github): gate POST /webhook on role and evict replay cache on upstream failure

yairfalse:
## Summary

Three bugs in the Phase 1 GitHub-native receiver (PR #125), surfaced by an independent code review of Codex-authored code. All three only manifest in production failure modes the existing test suite did not exercise.

### Critical bugs fixed

1. **`POST /webhook` was not gated by `Mesh.Roles.held_by_local?(:webhook_receiver)`** — only `GET /healthz` was. On a multi-node deployment every node would ingest deliveries, burn `delivery_id`s in their local replay caches, and create duplicate check suites/runs. POST now returns `503 inactive` when the local node does not hold the role, matching `/healthz`.

2. **Replay cache burned the delivery before upstream calls succeeded.** `accept_delivery` (which inserts the `delivery_id` into the replay cache) ran *before* `installation_token` / `create_suite` / `create_run`. A transient GitHub 5xx left the `delivery_id` in the cache; GitHub's automatic retry hit `:duplicate_delivery` and the webhook was permanently lost. Receiver now evicts the `delivery_id` on any failure that occurs after `accept_delivery`, so retries succeed. Concurrent dedup is preserved because `:ets.insert_new` in `Deliveries.accept/3` stays atomic and runs first.

3. **Unbounded `read_body/1`** accepted up to Plug's default 8 MB. Bounded to 1 MB with a 15s read timeout; oversize requests get a 413 with a distinct `github.webhook.body_too_large` error code.

### Higher-quality error responses

- **Missing vs. bad signature:** previously both returned 401 `bad_signature`. The missing-header case now returns 400 `github.webhook.missing_signature` so logs distinguish a stripped-header proxy bug from an actual forgery attempt.

- **502 instead of 400 for raw upstream failures:** the catch-all error branch returned 400 \"invalid GitHub webhook\" for non-`Sykli.Error` reasons, including raw `:httpc` upstream failures. GitHub does not retry 400s, so transient upstream failures looked like permanent client-side bad payloads. Catch-all now returns 502 `github.webhook.upstream_failure`, which GitHub will retry.

- **`Sykli.Mesh.Roles` moduledoc tightened** to make the local-only ETS semantics explicit. The \"mesh\" name was misleading — the registry is per-node single-holder enforcement, not cluster-coordinated. Multi-node deployments must ensure only one node carries a given role label; callers should gate on `held_by_local?/1` (which the receiver now does, see #1 above).

## Test plan

- [x] 1,436 tests pass (was 1,433 — added 3 new tests)
- [x] `mix credo` clean
- [x] `mix escript.build` succeeds
- [x] New tests added:
  - POST returns 503 when role not held
  - Missing signature header returns 400 with distinct code
  - Upstream Checks API failure evicts delivery → retry succeeds
- [x] Existing tests updated: acquire role in setup where POST is exercised; integration test \"rejects unsigned\" now expects 400 `missing_signature` instead of 401 `bad_signature`
- [ ] Manual verification in staging once Phase 2 webhook→executor wiring lands

## Out of scope

Three lower-priority findings from the same review are intentionally **not** in this PR:

- `Sykli.GitHub.HTTPClient` and `Sykli.GitHub.Checks` lack the `Behaviour`/`Real`/`Fake` split that `App` and `Clock` have — refactor, not a bug. Each test currently hand-rolls its own stub. Worth a separate PR before Phase 2 needs executor mocking.
- `Deliveries.prune/2` is O(n) `tab2list` per accept — performance, not correctness. Convert to `:ets.select_delete/2` with a match spec when load warrants it.
- `app/cache.ex` does not cap `expires_at` to `now + 3600` — defense in depth against a malformed GitHub response. Real-world risk is low.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,21 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **GitHub webhook receiver — three Phase 1 (ADR-021) correctness gaps.**
+  - `POST /webhook` is now gated by `Sykli.Mesh.Roles.held_by_local?(:webhook_receiver)`. Previously only `GET /healthz` was gated; on a multi-node deployment every node would ingest deliveries, burn `delivery_id`s in their local replay caches, and create duplicate check suites/runs.
+  - The replay cache no longer permanently loses deliveries when a downstream call fails. Previously, `accept_delivery` inserted the `delivery_id` *before* `installation_token` / `create_suite` / `create_run`, so a transient GitHub 5xx left the entry in cache and GitHub's automatic retry hit `:duplicate_delivery` / 409. The receiver now evicts the `delivery_id` on any post-accept failure; concurrent dedup is preserved by the atomic `:ets.insert_new` in `Deliveries.accept/3`.
+  - Request body is bounded to 10 MB (with a 15s read timeout) instead of Plug's default unbounded read. Distinct error codes for `body_too_large` (413) and `body_read_failed` (408).
+
+### Changed
+
+- **GitHub webhook error responses now distinguish missing vs. bad signature.** Missing `X-Hub-Signature-256` returns 400 `github.webhook.missing_signature`; an invalid signature still returns 401 `github.webhook.bad_signature`. Operators can now tell a stripped-header proxy bug apart from a forgery attempt.
+- **GitHub webhook catch-all is now 502 `github.webhook.upstream_failure`** for raw upstream errors (was 400 `github.webhook.invalid`). GitHub does not retry 400s, so the previous response misclassified transient upstream failures as permanent client errors and silently dropped them.
+- **`Sykli.Mesh.Roles` moduledoc** clarified to make the local-only ETS semantics explicit. The "mesh" name was misleading — the registry is per-node single-holder enforcement, not cluster-coordinated. Multi-node deployments must ensure only one node carries a given role label.
+
 ## [0.6.0] - 2026-04-29
 
 This release crystallizes the v0.6 line: a Nordic-minimal CLI, a fully decoupled runtime layer with deterministic-test defaults, SLSA v1.0 supply-chain attestations, an oracle-based AI agent evaluation harness, the FALSE-Protocol-first-class internal model, and the foundation for v0.7's GitHub-native CI integration. v0.5.1, v0.5.2, and v0.5.3 were development tags without CHANGELOG entries; their changes are folded in here.
--- a/core/lib/sykli/github/webhook/deliveries.ex
+++ b/core/lib/sykli/github/webhook/deliveries.ex
@@ -32,6 +32,23 @@ defmodule Sykli.GitHub.Webhook.Deliveries do
     :ok
   end
 
+  @doc """
+  Removes a delivery_id from the replay cache.
+
+  Used by the receiver to release a cache entry when downstream processing
+  failed after `accept/3` already inserted it, so GitHub's automatic retry
+  for the same delivery_id can succeed instead of hitting `:duplicate_delivery`.
+
+  Idempotent: deleting a missing key is a no-op.
+  """
+  def evict(delivery_id) when is_binary(delivery_id) do
+    ensure_table()
+    :ets.delete(@table, delivery_id)
+    :ok
+  end
+
+  def evict(_other), do: :ok
+
   defp prune(now_ms, ttl_ms) do
     threshold = now_ms - ttl_ms
 
--- a/core/lib/sykli/github/webhook/receiver.ex
+++ b/core/lib/sykli/github/webhook/receiver.ex
@@ -9,7 +9,10 @@ defmodule Sykli.GitHub.Webhook.Receiver do
   alias Sykli.Occurrence.PubSub, as: OccPubSub
 
   @role :webhook_receiver
-  @max_body_bytes 1_000_000
+  # GitHub's documented webhook payload ceiling is 25 MB; 10 MB covers
+  # large-PR `pull_request` and many-commit `push` payloads with headroom
+  # while still bounding worst-case memory per request.
+  @max_body_bytes 10_000_000
   @read_timeout_ms 15_000
 
   def init(opts), do: opts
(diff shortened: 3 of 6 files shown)


yairfalse (commented):
## Code Review

### Overview

Three production-only correctness fixes to the Phase 1 receiver, plus error-code refinements:

1. Role gate on `POST /webhook` (was missing; only `GET /healthz` was gated).
2. Replay-cache eviction on post-accept failure (transient GitHub 5xx was permanently losing webhooks).
3. Bounded request body (1 MB + 15s read timeout) replacing Plug's default 8 MB unbounded `read_body/1`.

Plus: distinct `missing_signature` (400) vs `bad_signature` (401), `502 upstream_failure` catch-all instead of 400, tightened `Sykli.Mesh.Roles` moduledoc.

### Strengths

- **Failure-mode reasoning is correct.** Insert atomically via `:ets.insert_new`, evict deterministically post-accept. The eviction-then-retry test (`receiver_test.exs:176-207`) is a great regression guard.
- **Status codes match retry semantics.** 502 → GitHub retries; 401/400 → it doesn't. The catch-all 502 is the most operationally important fix in the PR — silently misclassifying upstream 5xx as 400 was turning transient failures into permanent drops.
- **`evict/1` design is sound.** Idempotent, accepts non-binary defensively, and the receiver-side wrapper handles missing-header cleanly.
- **`Mesh.Roles` doc clarification** is well-placed — the \"mesh\" name was actively misleading; future engineers could easily have assumed cluster-wide coordination.

### Issues

#### 🔴 Body size limit is too tight for real GitHub payloads

`@max_body_bytes 1_000_000` (1 MB) in `receiver.ex:12`. GitHub's documented webhook ceiling is 25 MB. Real-world `pull_request` events with file diffs and `push` events with many commits routinely exceed 1 MB. As shipped, this PR will return 413 on legitimate webhooks and lose them (GitHub does not retry 413).

Recommend bumping to ~10 MB and exposing it via `opts` so the 413 path is testable without allocating a real oversized body. A 413 test should ship with the change — currently the body-too-large path is untested.

#### 🟡 Missing CHANGELOG entry

User-visible bug fix in the GitHub-native foundation; the project follows Keep-a-Changelog (per CLAUDE.md). Add an entry — especially for the \"delivery_id was being burned on transient failures\" item, since it changes operator-observable behavior.

#### 🟡 Eviction is over-broad on 4xx errors

`process_accepted/3` returns errors for `decode_json` (400), `webhook_context` (400 `unsupported_payload`), and the upstream calls (502). Current code evicts on all of them. For 4xx cases this is harmless but pointless — GitHub will not retry, so eviction is wasted. A future reader may wonder if it's load-bearing. Either:

- Add a one-line comment noting \"evict unconditionally; classifying 4xx vs 5xx isn't worth the complexity, and GitHub retrying a malformed body is improbable\", **or**
- Restrict eviction to the non-`Sykli.Error` clause of `respond_error/2`.

First option is probably right — just document it.

#### 🟢 Minor: `verify_signature` cond ordering

`is_nil(secret)` is checked before `is_nil(signature)`, so a server with no secret configured AND a request with no signature gets `503 missing_secret` rather than `400 missing_signature`. This is correct (server misconfig wins) but worth a one-line comment to keep a future edit from flipping the order.

#### 🟢 Minor: `body_read_failed` → 400 is debatable

`status_for/1` maps `body_read_failed` to 400 (`receiver.ex:263`). This branch fires on `read_body/1` returning `{:error, reason}` — typically a network/IO/timeout on the client connection, not a malformed request. 408 or 502 is more honest than 400. Low priority.

### Test Coverage

- ✅ POST without role → 503
- ✅ Missing signature → 400 distinct code
- ✅ Upstream 5xx → 502 + eviction → retry succeeds
- ❌ **Body too large → 413** — should ship with this PR
- ❌ Behavior on `decode_json` / `unsupported_payload` failures (does eviction fire? — would document the design decision)

Role-acquisition pattern in tests (every POST-exercising test calls `Roles.acquire(:webhook_receiver)`; setup `Roles.clear()`s) is correct, and `async: false` is necessary due to the shared ETS table.

### Security

- HMAC verification still runs before any side effects. ✅
- Replay-cache eviction does not weaken dedup: `:ets.insert_new` remains the atomic gate; eviction only fires after a request has finished failing, so two concurrent identical deliveries cannot both proceed. ✅
- Body size cap closes the trivial DoS via large POSTs ✅ — at the wrong value, but the mechanism is right.
- Missing-signature → 400 is debatable vs strict HTTP semantics (some readers expect 401 for any auth failure), but the diagnostic-distinction argument is reasonable and the body still carries a structured `error.code`.

### Recommendation

**Approve with one blocking change**: bump `@max_body_bytes` from 1 MB to ~10 MB (with a `:max_body_bytes` opt override for tests) and add the 413 test before merging. Other items are nits or follow-ups.

The three out-of-scope items in the PR description (`HTTPClient`/`Checks` behaviour split, `prune/2` O(n), `expires_at` clamping) are correctly punted — none are correctness bugs.

yairfalse:
Thanks for the thorough review. Quick reconciliation against what's actually on the PR head — there's a timeline ambiguity that's worth clearing up:

**Already fixed in `b6a54fa`** (pushed ~3 min before your review submitted, so you may have been drafting against the original commit):

- 🔴 `@max_body_bytes` 1 MB → **10 MB** — now declared at `receiver.ex:12-15` with a comment citing GitHub's 25 MB ceiling.
- ❌ **413 test** — see `receiver_test.exs` "request body over the configured size limit returns 413". Uses a `:max_body_bytes` opt override (1-line code seam) so the test exercises the path with 10 bytes instead of 10 MB of test memory.

**Just landed in `3826c88`** in response to this review:

- 🟡 **CHANGELOG entry** — added `[Unreleased] / Fixed / Changed` section covering the three correctness fixes plus the response-code refinements.
- 🟡 **Comment on the eviction-on-any-error choice** — added at `receiver.ex` near the `evict_delivery` call, explicitly noting that for 5xx eviction is load-bearing (lets GitHub retry succeed) and for 4xx it's a no-op in practice (GitHub doesn't retry malformed payloads), so classifying retryable-vs-terminal would add code without changing behavior.
- 🟢 **`verify_signature` cond order comment** — locks in `missing_secret` (503) winning over `missing_signature` (400) so a future refactor doesn't flip them and make every request to a misconfigured server look like a client problem.
- 🟢 **`body_read_failed`: 400 → 408** — agreed, the error fires on Plug's `read_body` returning `{:error, _}` which is a timeout/IO error, not a malformed request. 408 is more honest.

**Not adding** (with rationale):

- `try/rescue` around `process_accepted` for hypothetical raised exceptions — project policy (CLAUDE.md: *"Don't add error handling, fallbacks, or validation for scenarios that can't happen"*). No callee in the chain raises today.
- Test for `decode_json` / `unsupported_payload` failures firing eviction — the new comment documents the design; a test that confirms what the comment says would be redundant.

PR head: `3826c88`. 1,439 tests passing, `mix credo` clean.

PR #142, opened 2026-05-02
feat(github): Phase 2 — Webhook → Executor → Check Run Lifecycle (ADR-021)

yairfalse:
Summary: closes the GitHub-native loop from accepted webhook to local execution and per-task GitHub check runs.\n\nScope:\n- Adds Source behaviour with Real git clone and Fake fixture-backed source acquisition.\n- Splits GitHub Checks behind Behaviour/Real/Fake and adds HTTPClient.Fake.\n- Adds Dispatcher to acquire source, build the graph, run Executor, and transition per-task check runs.\n- Wires Receiver to return 202 after accepted webhook and dispatch asynchronously under Sykli.TaskSupervisor.\n- Adds Phase 2 GitHub-native occurrence types and check-run formatting.\n- Updates docs/github-native.md for contents:read, per-task runs, and new occurrences.\n\nVerification:\n- cd core && mix test\n- cd core && mix credo\n- cd core && mix escript.build\n- test/blackbox/run.sh: 137 passed, 11 expected-red, 0 failed\n\nScreenshot/API proof:\n- Not attached: this PR uses local fakes and fixture source for the Phase 2 loop. A live PR Checks-tab capture requires an installed GitHub App on an exposed receiver endpoint.

### DIFF
--- a/core/lib/sykli/github/app/fake.ex
+++ b/core/lib/sykli/github/app/fake.ex
@@ -5,8 +5,14 @@ defmodule Sykli.GitHub.App.Fake do
 
   @impl true
   def installation_token(installation_id, opts \\ []) do
-    token = Keyword.get(opts, :token, "fake-installation-token-#{installation_id}")
-    expires_at = Keyword.get(opts, :expires_at, 4_102_444_800)
-    {:ok, token, expires_at}
+    case Keyword.get(opts, :app_response, :default) do
+      :default ->
+        token = Keyword.get(opts, :token, "fake-installation-token-#{installation_id}")
+        expires_at = Keyword.get(opts, :expires_at, 4_102_444_800)
+        {:ok, token, expires_at}
+
+      response ->
+        response
+    end
   end
 end
--- a/core/lib/sykli/github/check_run_formatter.ex
+++ b/core/lib/sykli/github/check_run_formatter.ex
@@ -1,0 +1,73 @@
+defmodule Sykli.GitHub.CheckRunFormatter do
+  @moduledoc "Formats task results for GitHub check runs."
+
+  alias Sykli.Executor.TaskResult
+
+  @max_lines 50
+
+  @spec conclusion(TaskResult.t()) :: String.t()
+  def conclusion(%TaskResult{status: :passed}), do: "success"
+  def conclusion(%TaskResult{status: :failed}), do: "failure"
+  def conclusion(%TaskResult{status: :errored}), do: "failure"
+  def conclusion(%TaskResult{status: :cached}), do: "success"
+  def conclusion(%TaskResult{status: :skipped}), do: "skipped"
+  def conclusion(%TaskResult{status: :blocked}), do: "cancelled"
+
+  @spec format(TaskResult.t()) :: %{title: String.t(), summary: String.t()}
+  def format(%TaskResult{} = result) do
+    %{
+      title: title(result),
+      summary: summary(result)
+    }
+  end
+
+  defp title(%TaskResult{name: name, status: :cached}), do: "#{name}: cached (cache hit)"
+  defp title(%TaskResult{name: name, status: status}), do: "#{name}: #{status}"
+
+  defp summary(%TaskResult{status: :errored} = result) do
+    [
+      "Infrastructure failure while running `#{result.name}`.",
+      "",
+      output_block(result)
+    ]
+    |> Enum.reject(&(&1 == ""))
+    |> Enum.join("\n")
+  end
+
+  defp summary(%TaskResult{} = result) do
+    [
+      "Task `#{result.name}` finished with status `#{result.status}`.",
+      output_block(result)
+    ]
+    |> Enum.reject(&(&1 == ""))
+    |> Enum.join("\n\n")
+  end
+
+  defp output_block(%TaskResult{} = result) do
+    output =
+      [result.output, error_output(result.error)]
+      |> Enum.reject(&is_nil/1)
+      |> Enum.join("\n")
+      |> last_lines(@max_lines)
+
+    if output == "" do
+      ""
+    else
+      "```text\n#{output}\n```"
+    end
+  end
+
+  defp error_output(%Sykli.Error{output: output}) when is_binary(output), do: output
+  defp error_output(%Sykli.Error{message: message}) when is_binary(message), do: message
+  defp error_output(nil), do: nil
+  defp error_output(error), do: inspect(error)
+
+  defp last_lines("", _limit), do: ""
+
+  defp last_lines(text, limit) do
+    text
+    |> String.split("\n")
+    |> Enum.take(-limit)
+    |> Enum.join("\n")
+  end
+end
--- a/core/lib/sykli/github/checks.ex
+++ b/core/lib/sykli/github/checks.ex
@@ -1,113 +1,39 @@
 defmodule Sykli.GitHub.Checks do
-  @moduledoc "GitHub Checks API client."
+  @moduledoc "GitHub Checks API facade."
 
-  require Logger
+  @behaviour Sykli.GitHub.Checks.Behaviour
 
-  @api_url "https://api.github.com"
+  @impl true
+  def create_suite(context, token, opts \\ []) do
+    impl =
+      Keyword.get(opts, :impl, Application.get_env(:sykli, :github_checks_impl, __MODULE__.Real))
 
-  def create_suite(%{repo: repo, head_sha: head_sha}, token, opts \\ []) do
-    request(
-      :post,
-      "/repos/#{repo}/check-suites",
-      token,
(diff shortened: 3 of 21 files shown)


yairfalse (commented):
## Code Review

### Overview

Closes the GitHub-native loop: receiver → async dispatch → source clone → executor → per-task Check Runs. Adds 8 new modules: the Checks `Behaviour/Real/Fake` split, `HTTPClient.Fake`, `Source.{Behaviour,Real,Fake}`, `Dispatcher`, and `CheckRunFormatter`. Receiver becomes 202-fast (offloads to `Task.Supervisor`). Status mapping in `CheckRunFormatter` matches the ADR-021 contract exactly. 7 new occurrence types wired through factory + PubSub. End-to-end test in `dispatcher_test.exs` covers happy path and clone-failure-evicts-delivery.

The structural shape is correct. Flagging issues by severity.

### 🔴 Critical

**Installation token leaks via `git clone` URL.** `source/real.ex:1026` builds `https://x-access-token:<token>@github.com/<repo>.git` and passes it as a CLI arg in `source/real.ex:983`. Two real exposure paths:

1. **Process listing.** On macOS `ps -E` shows full cmdline for all users by default; on Linux `/proc/<pid>/cmdline` is readable by the same UID. While clones are short-lived (~seconds), any concurrent process at the right moment captures the token.
2. **`.git/config` persistence.** After clone, `.git/config` contains `[remote \"origin\"] url = https://x-access-token:TOKEN@github.com/...`. The token is now on disk inside the cloned workspace until cleanup. The executor's tasks run inside that workspace and can read `.git/config`.

Both are industry-standard concerns for token-in-URL git auth. Fix options, in order of preference:
- `git -c http.extraheader=\"Authorization: Bearer <token>\" clone https://github.com/<repo>.git ...` — token never enters argv or `.git/config`.
- Pipe credentials via `GIT_ASKPASS` script + env var.
- After clone, run `git remote set-url origin https://github.com/<repo>.git` to scrub the persisted URL (mitigates #2 but not #1).

The output masking via `SecretMasker.mask_string(out, [token])` in `clone/3` is good but only protects the error report — the token is still in argv and `.git/config` regardless of whether stderr is captured.

### 🟠 High

**`try/after` cleanup is not crash-safe** (`dispatcher.ex:498-500`). If the dispatch task is killed externally (supervisor restart, `:exit`, OOM, node shutdown), `after` does not fire and the workspace leaks. The original Phase 2 design called for process-monitor cleanup — a separate process holding the workspace path that triggers `File.rm_rf` on the dispatcher's `:DOWN` message. Over time on a long-running mesh, leaked workspaces accumulate under `System.tmp_dir!()/sykli-runs/`. Suggest filing as a follow-up if not fixing in this PR; at minimum, a periodic janitor (e.g., on dispatcher startup, scan `@base_dir` for entries older than N hours and `rm_rf`).

**`File.rm_rf!(root) |> always_ok()` will crash the dispatch task on permission errors** (`source/real.ex:946`). The `!` variant raises `File.Error`. The `with` clause expects `:ok | {:error, _}` — exceptions propagate past it, past `do_dispatch`, into the supervised task which dies. Workspace from a prior run with weird perms (e.g., a sticky `.git/objects` from a crashed `git clone`) will reliably brick fresh runs. Use non-bang `File.rm_rf/1` and either ignore the result or pattern-match.

### 🟡 Medium

**Phase 1 \"queued-only\" degraded mode is not preserved.** Prompt §\"Don't break Phase 1's queued-only path on degraded configs\" asked for: missing App creds → still 202, still emit `webhook.received`, log structured warning, stop. Current behavior: receiver returns 202 (good), dispatcher tries `installation_token`, gets `{:error, _}`, **evicts the delivery**, GitHub retries, fails again. The eviction is correct for transient infra failures but wrong for permanent config errors — endless retry storm. Distinguish: clone/Checks-API errors → evict (transient); App-credential errors → no evict, log, give up. Or: gate the entire dispatch on `installation_token` being available before declaring \"this delivery is dispatchable.\"

**`run_executor_quietly` swaps the process group leader to silence executor output** (`dispatcher.ex:624-636`). This is brittle. The Executor already routes user-facing output through `Sykli.Executor.Output` — a configurable sink would be cleaner. Group-leader manipulation also affects any downstream `IO.puts` from inside tasks (which a runtime might call), making per-task `output` capture potentially incomplete. Worth tracking as a refactor: add `:silent` or `:io_device` opt to `Executor.run/3`.

**`mark_in_progress` discards all transition errors** (`dispatcher.ex:583-601`). `transition_check_run` only logs a warning on failure, and `mark_in_progress` always returns `:ok`. If GitHub rate-limits the burst of N PATCH calls, every Check Run stays at `queued` forever (until the final completion update fires). The `transition_failed` occurrence is emitted but nothing acts on it. Consider: aggregate failures and abort the run if ≥50% of in-progress transitions fail (likely a token or rate-limit issue worth bailing on).

**`directory_bytes` walks the entire cloned tree including `.git/`** (`dispatcher.ex:750-761) for one field on one occurrence. For a 50k-file monorepo this can be hundreds of milliseconds per webhook. Either drop the field or replace with a cheap `du -sh` shell-out. Currently this fires on every successful clone.

**Two API calls per task to start** (`mark_in_progress` + `conclude_task_runs`). GitHub Apps have 5000 req/hr per installation. A 30-task pipeline = 60 calls just for Check Runs (create + transition + complete). Acceptable for now but worth flagging as the load grows. Could create directly at `in_progress` and skip `mark_in_progress` entirely (the brief `queued` window has minimal UX value — the user sees the suite for milliseconds before tasks start).

### 🟢 Low / nits

- `:source_client` (Dispatcher's opts key) vs `:impl` (Source.acquire's opts key) — same concept, different names. Dispatcher reaches in via `source_client`, public Source API uses `impl`. Pick one.
- `safe_segment` regex preserves `:` (`source/real.ex:1060`, `source/fake.ex:901`). Fine on darwin/linux, illegal in Windows path components. Sykli is Unix-targeted so low priority, but worth a one-line comment if you want to keep the colon.
- `source_test.exs` test \"real cleanup refuses paths outside the sykli temp root\" only asserts the return value is `:ok`. Doesn't verify the path was preserved. Stronger assertion: create `/tmp/not-sykli-source-decoy/sentinel`, call cleanup, assert the sentinel still exists.
- `dispatcher.ex:728` suite-level `conclusion/1` lumps `:blocked` into `failure`, but `CheckRunFormatter` maps `:blocked` → `cancelled`. Per the ADR-021 prompt, GitHub computes the suite conclusion from the runs anyway, so the dispatcher's emitted suite-conclusion in the occurrence is for downstream consumers only. Either drop the calculation or align with the formatter table.
- Receiver test now uses `eventually` polling for the retry case (`receiver_test.exs:1645-1675`). Bounded at 200ms, probably fine, but susceptible to slow CI. Stronger: wait for an explicit `:dispatch_completed` signal from the mock dispatcher.

### Test Coverage

- ✅ Status mapping table — full coverage in `check_run_formatter_test.exs`.
- ✅ Happy-path dispatch creates suite, per-task runs, and transitions — `dispatcher_test.exs`.
- ✅ Clone-failure evicts delivery — `dispatcher_test.exs`.
- ✅ Receiver returns 202 immediately and hands off to dispatcher — `receiver_test.exs`.
- ❌ Token-in-URL test (verify token does not appear in argv, `.git/config`, or any persisted file inside the workspace).
- ❌ Workspace leak test (kill the dispatch task mid-clone, assert workspace is cleaned up — fails today, would document the `try/after` gap).
- ❌ Degraded-mode test (missing App credentials → no infinite retry storm).
- ❌ `directory_bytes` perf bound (synthetic 10k-file fixture, assert dispatch < N ms).
- ❌ Concurrent dispatch (two webhooks for the same SHA, different `delivery_id`, hitting `@base_dir` simultaneously — `safe_segment(run_id)` should isolate, but no test confirms).

### Security Summary

- ✅ HMAC verification preserved in receiver.
- ✅ Replay-cache eviction preserved (PR #141 contract).
- ✅ Path containment in `Source.Real` (`ensure_contained`, `run_root` reject non-base paths).
- ✅ Output masking via `SecretMasker.mask_string` in clone/checkout error reports.
- ❌ **Token leaks via argv and `.git/config`** (see Critical).
- ⚠ Cleanup-on-process-crash gap (see High).

### Recommendation

**Hold for the token-in-URL fix.** The other items can be PR follow-ups, but argv/`.git/config` token exposure is a real security regression vs Phase 1 (which never put a token on the filesystem). Fix is mechanical — `git -c http.extraheader=...` is one line. Worth landing in this PR alongside a test that asserts the token does not appear in the workspace's `.git/config`.

The structural work (Behaviour split, dispatcher composition, occurrence types, receiver wiring) is in good shape — once the security issue is addressed, this is mergeable with the Medium items as tracked follow-ups.

PR #151, opened 2026-05-03
chore(release): add deterministic release automation

yairfalse:
Summary:
- Adds root VERSION as the release source of truth.
- Replaces the version bump helper with checked, dry-runnable manifest updates across core + SDKs.
- Adds release, publish-all, and per-registry publish scripts plus Makefile targets.
- Updates RELEASE.md with the operator flow, credentials, and example dry-run output.

Validation:
- bash -n scripts/*.sh
- scripts/check-version.sh
- scripts/bump-version.sh 0.6.1 && scripts/bump-version.sh 0.6.0
- make dry-run VERSION=0.6.0
- make publish VERSION=0.6.0 fails before publishing with missing credential errors
- cd core && mix test

Conformance note:
- tests/conformance/run.sh was started and stopped after surfacing the existing SDK parity failures (Go outputs:null mismatches plus Python/TS/Rust/Elixir execution failures). The new release script enforces this gate for actual releases, so a real release will fail until conformance is green.


scripts/release.sh
```diff
@@ -0,0 +1,119 @@
...
+    log "dry run: actual release would fail because tag already exists: v$VERSION"
+  else
+    die "tag already exists: v$VERSION"
+  fi
+fi
+
+log "checking version consistency before bump"
+run "$ROOT/scripts/check-version.sh"
+
+log "running core tests"
```
yairfalse, line 87: **Test ordering: tests run before the bump.**

The full test gate (lines 78–89) runs against the *pre-bump* version strings. The bump happens on line 91, then commit + tag on 105/107. The tag therefore points at a commit whose bumped version strings were never tested.

For sykli specifically the practical risk is low (version isn't load-bearing at runtime), but the script's whole purpose is rigor — "tested code = tagged code" is the rigor that matters most.

**Suggested fix:** bump first, then run tests, then commit + tag. If tests fail post-bump, recovery is `git checkout -- .`.

scripts/publish-python.sh
```diff
@@ -0,0 +1,34 @@
...
+  exit 0
+fi
+
+if [[ -z "${PYPI_API_TOKEN:-}" && -z "${TWINE_PASSWORD:-}" ]]; then
+  echo "[publish-python] error: PYPI_API_TOKEN or TWINE_PASSWORD is required" >&2
+  exit 1
+fi
+
+cd "$(dirname "${BASH_SOURCE[0]}")/../sdk/python"
+python3 -m build
```
yairfalse, line 35: **Should clean `dist/` before build.**

If a stale wheel from a prior aborted publish is in `dist/`, `twine upload dist/*` on line 31/33 will try to upload it too — best case a duplicate-version error from PyPI, worst case publishing the wrong artifact.

**Suggested fix:** `rm -rf dist/ build/ *.egg-info/` immediately before `python3 -m build`.

scripts/publish-all.sh
```diff
@@ -0,0 +1,43 @@
...
+args=()
+if [[ "$DRY_RUN" -eq 1 ]]; then
+  args+=(--dry-run)
+fi
+
+"$ROOT/scripts/publish-go.sh" "$VERSION" "${args[@]}"
+"$ROOT/scripts/publish-rust.sh" "$VERSION" "${args[@]}"
+"$ROOT/scripts/publish-ts.sh" "$VERSION" "${args[@]}"
+"$ROOT/scripts/publish-python.sh" "$VERSION" "${args[@]}"
+"$ROOT/scripts/publish-elixir.sh" "$VERSION" "${args[@]}"
```
yairfalse: **No rollback or resume semantics for partial publish failures.**

If `publish-go.sh` succeeds (tag pushed) and then `publish-rust.sh` fails (e.g. network blip mid-`cargo publish`), you're in a half-published state. crates.io and Go modules have very different rollback semantics — atomic multi-registry publishing isn't really possible.

**Minimum mitigation:**
- Document the recovery procedure in `RELEASE.md`: "if publish fails mid-flight, the failing script is idempotent — rerun it with the same version, then continue manually with the remaining publishers."
- Audit each publisher to confirm idempotency under partial state (`publish-go.sh` already handles existing local tag — the others should be checked).

**Bonus:** consider ordering by reversibility — Go module tag is the most reversible (just delete the tag), npm/crates.io are essentially irreversible (yank ≠ delete). Publishing the irreversible ones last gives the most chance to abort mid-flight.

scripts/release.sh
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env bash
```
yairfalse, line 1: **No tests for the release scripts themselves.**

Validation is manual (`scripts/bump-version.sh 0.6.1 && scripts/bump-version.sh 0.6.0` per the PR description). For high-stakes infrequently-run scripts, manual is the realistic baseline — but a small smoke test (~30 LOC under `tests/release/`) that fixtures a manifest layout, runs `bump-version.sh` against it, and diffs the result would catch regressions when these scripts are touched in 6 months by someone who doesn't remember the patterns.

`bats` or plain bash + `diff` would do; doesn't need ceremony.

scripts/release.sh
```diff
@@ -0,0 +1,119 @@
...
+
+if [[ -n "$(git status --porcelain)" ]]; then
+  if [[ "$DRY_RUN" -eq 1 ]]; then
+    log "dry run: actual release would fail because the working tree is not clean"
+  else
+    die "working tree is not clean"
+  fi
+fi
+
+if git rev-parse "v$VERSION" >/dev/null 2>&1; then
```
yairfalse, line 70: **Local-only tag check.** `git rev-parse "v$VERSION"` only checks local tags. A tag could exist upstream but not locally — the release would proceed and fail at push time, after work has been done.

**Suggested fix:** `git fetch --tags` immediately before this check.

scripts/release.sh
```diff
@@ -0,0 +1,119 @@
...
+
+if ! [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+([.-][0-9A-Za-z.-]+)?$ ]]; then
+  die "version must be semver without a leading v, got '$VERSION'"
+fi
+
+cd "$ROOT"
+
+log "release v$VERSION"
+
+if [[ -n "$(git status --porcelain)" ]]; then
```
yairfalse, line 59: **`git status --porcelain` includes untracked files.** A `.DS_Store`, IDE config, or local scratch file will block the release even when the tracked tree is clean.

Consider `git diff --quiet && git diff --cached --quiet` instead, allowing untracked files. Strict-clean is defensible (and simpler) — just flagging the trade-off.

scripts/release.sh
```diff
@@ -0,0 +1,119 @@
...
+  log "would commit release: v$VERSION"
+  log "would tag v$VERSION"
+else
+  git add VERSION core/mix.exs sdk/elixir/mix.exs sdk/rust/Cargo.toml sdk/python/pyproject.toml sdk/typescript/package.json sdk/typescript/package-lock.json
+  if git diff --cached --quiet; then
+    log "no version files changed; release commit is not needed"
+  else
+    git commit -m "release: v$VERSION"
+  fi
+  git tag -a "v$VERSION" -m "v$VERSION"
```
yairfalse, line 110: **Unsigned tag.** `git tag -a` creates an annotated but unsigned tag. Sykli already produces SLSA attestations for build artifacts — signed tags (`git tag -s`) would be the matching rigor for releases, giving downstream verifiers a cryptographic provenance chain from tag → artifact.

Not blocking, but worth queuing as a follow-up given the project's existing supply-chain investment.

scripts/publish-elixir.sh
```diff
@@ -0,0 +1,29 @@
...
+  exit 0
+fi
+
+if [[ -z "${HEX_API_KEY:-}" ]]; then
+  echo "[publish-elixir] error: HEX_API_KEY is required" >&2
+  exit 1
+fi
+
+cd "$(dirname "${BASH_SOURCE[0]}")/../sdk/elixir"
+mix hex.publish --yes
```
yairfalse, line 29: **`mix hex.publish --yes` bypasses interactive confirmation.** Acceptable in automation, and the upstream `release.sh` test gate mitigates — but worth noting that an upstream bug here could publish anything that passes the gate. No change requested; flagging for awareness.

scripts/publish-python.sh
```diff
@@ -0,0 +1,34 @@
...
+if [[ -n "${PYPI_API_TOKEN:-}" ]]; then
```
yairfalse, line 36: **`TWINE_PASSWORD` fallback inherits ambient env.** If `PYPI_API_TOKEN` is unset but `TWINE_PASSWORD` is set, twine uses whatever `TWINE_USERNAME` is in the environment — which could be anything (a leftover from another project, the user's actual PyPI username, unset entirely).

The credential check in `publish-all.sh:25` requires one of them to be set, but doesn't pin the username. Consider erroring here if `TWINE_USERNAME` isn't also explicitly set when going down the fallback path.

scripts/publish-go.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
```
yairfalse, line 1: **Flag-parsing duplication across all 5 publish scripts.** The `VERSION="${1:-}" / DRY_RUN=0 / shift / for arg in $@ / case` block is repeated identically in `publish-{go,rust,ts,python,elixir}.sh`.

A `scripts/lib/parse-publish-args.sh` sourced helper would DRY this and prevent drift when one script's flag handling diverges from the others. Not blocking; cleanup pass.

RELEASE.md
```diff
@@ -1,140 +1,76 @@
-# Sykli v1.0.0 Release
+# Release Automation
```
yairfalse, line 1: **Missing prereq:** the new bump/check scripts shell out to `python3` (heredoc-via-env-var pattern). Worth a one-line note in RELEASE.md that `python3` is required for the version-management scripts — macOS ships it and most Linux dev boxes have it, but the *old* sed-based scripts didn't need it, so an operator following an old runbook could be surprised.

Suggest adding under a new "Prerequisites" section: `python3` (for version manifest management) and `git` ≥ 2.x.

Makefile
```diff
@@ -0,0 +1,18 @@
+VERSION ?= $(shell tr -d '[:space:]' < VERSION)
```
yairfalse, line 1: **Nit:** `VERSION ?= $(shell tr -d '[:space:]' < VERSION)` evaluates on every Make invocation, even for targets that don't read VERSION (none today, but `bump-version` reads it via the script). Cost is negligible and the simplicity is worth more than the optimization — flagging only because it's the kind of thing that pile-of-Makefile-gotchas later. No action needed.

yairfalse (commented):
## Review summary

Solid release tooling. The Python-via-heredoc pattern in `bump-version.sh` / `check-version.sh` is genuinely thoughtful — sidesteps the macOS-vs-GNU `sed -i` incompatibility, supports proper named-group backreferences, and the `re.subn` count assertions catch "pattern matched too few/too many times" — a class of bug that silent `sed` swallows. Worth keeping as a project pattern (the `replace_all_checked` for `package-lock.json` with `expected_count=2` is a particularly nice touch — npm writes the version twice in lockfiles).

The PR description is exemplary: explicit about what was validated, what's known broken (pre-existing conformance failures), and that the new flow gates on conformance so a real release will fail until it's fixed. That's the right honesty.

### Real concerns (worth fixing before merge)

1. **Test ordering** — tests run before the bump, so the tagged commit's bumped versions are never tested. See inline on `release.sh:78`.
2. **`publish-python.sh` doesn't clean `dist/` before build** — stale wheels from prior aborted publishes would be uploaded. See inline on `publish-python.sh:29`.
3. **No rollback / resume semantics for partial publish failures** — if `publish-rust.sh` fails after `publish-go.sh` succeeds, you're in a half-published state with no defined recovery. See inline on `publish-all.sh:43`.
4. **No tests for the release scripts themselves** — manual smoke testing only. See inline on `release.sh:1`.

Items 5–12 are inline as smaller findings.

### What's good

- Dry-run mode is comprehensive across every script
- Credential checks happen before any network call (`publish-all.sh:21-32`)
- Version regex is consistently enforced and tight enough to prevent shell injection (`^[0-9]+\.[0-9]+\.[0-9]+([.-][0-9A-Za-z.-]+)?$`)
- `check-version.sh` runs both pre- and post-bump in `release.sh` — defense in depth against partial bumps
- Per-script `[name]` log prefixes are consistent

### Recommendation

Approve with the four real concerns addressed. 5–12 are nice-to-haves that can land separately.

scripts/release.sh
yairfalse: Fixed in ce43e91. `release.sh` now bumps and re-checks versions before running core/SDK/conformance tests, so the eventual release commit/tag corresponds to the tested versioned tree.

scripts/publish-python.sh
yairfalse: Fixed in ce43e91. `publish-python.sh` now removes `dist/`, `build/`, and `*.egg-info/` immediately before building.

scripts/publish-all.sh
yairfalse: Mitigated in ce43e91. `RELEASE.md` now documents partial-publish recovery and `publish-all.sh` supports `--from <go|rust|ts|python|elixir>` so operators can resume from the failed publisher without replaying already-published registries.

scripts/release.sh
yairfalse: Fixed in ce43e91. `release.sh` now fetches tags before checking whether `v$VERSION` exists, so it catches remote tags before bump/test/commit work.

scripts/release.sh
yairfalse: Fixed in ce43e91. Added `tests/release/run.sh` plus `make test-release-scripts`; it fixtures a minimal repo layout, runs check/bump/check, and asserts every manifest changed to the requested version.

scripts/publish-python.sh
yairfalse: Fixed in ce43e91. Both `publish-all.sh` preflight and `publish-python.sh` now require `TWINE_USERNAME` when using `TWINE_PASSWORD` without `PYPI_API_TOKEN`.

RELEASE.md
yairfalse: Fixed in ce43e91. Added a prerequisites section to `RELEASE.md` covering `git` 2.x, `python3`, and the language toolchains used by the release gate.

scripts/release.sh
yairfalse: Leaving this as a follow-up. I agree signed tags match the supply-chain direction, but switching release tags to `git tag -s` needs operator GPG/signing-key setup documented separately.

scripts/publish-go.sh
yairfalse: Leaving this as follow-up cleanup. The parser duplication is contained and the current patch stayed focused on correctness issues from this review.

PR #159, opened 2026-05-03
fix(github): avoid retry storm on App auth failures

yairfalse:
Summary:
- Treat all github.app.* dispatcher failures as permanent/degraded for replay-cache purposes.
- Do not evict the webhook delivery on App auth/config failures, preventing GitHub from retrying a doomed webhook forever.
- Keep eviction behavior for source/checks/executor failures so transient failures can still retry.
- Adds regression coverage for github.app.unauthorized.

Closes #152.

Validation:
- cd core && mix format
- cd core && env -u SYKLI_LABELS mix test was run; it hit an existing unrelated StreamData TooManyDuplicatesError in Sykli.Mesh.Transport.Sim.EventQueueTest.

### DIFF
--- a/core/lib/sykli/github/dispatcher.ex
+++ b/core/lib/sykli/github/dispatcher.ex
@@ -21,7 +21,9 @@ defmodule Sykli.GitHub.Dispatcher do
           Deliveries.evict(delivery_id)
           Logger.warning("[GitHub Dispatcher] dispatch failed", code: error.code)
         else
-          Logger.warning("[GitHub Dispatcher] degraded mode - App unavailable", code: error.code)
+          Logger.warning("[GitHub Dispatcher] App auth failed; delivery will not be retried",
+            code: error.code
+          )
         end
 
         result
@@ -54,13 +56,13 @@ defmodule Sykli.GitHub.Dispatcher do
   defp dispatch_after_suite(event, token, suite, opts) do
     run_id = event.run_id
 
-    with {:ok, source_path} <- acquire_source(event, token, opts),
-         {:ok, results} <- dispatch_from_source(event, token, source_path, opts) do
+    with {:ok, source_path, janitor} <- acquire_source(event, token, opts),
+         {:ok, results} <- dispatch_from_source(event, token, source_path, janitor, opts) do
       OccPubSub.github_check_suite_concluded(run_id, %{
         repo: event.repo,
         head_sha: event.head_sha,
         check_suite_id: suite["id"],
-        conclusion: conclusion(results)
+        conclusion: suite_conclusion(results)
       })
 
       :ok
@@ -94,16 +96,33 @@ defmodule Sykli.GitHub.Dispatcher do
     end
   end
 
-  defp dispatch_from_source(event, token, source_path, opts) do
-    with {:ok, graph, tasks} <- load_graph(source_path),
-         {:ok, check_runs} <- create_task_runs(event, token, tasks, opts),
-         :ok <- mark_in_progress(event, token, check_runs, opts),
-         {:ok, results} <- run_executor(tasks, graph, source_path, event.run_id, opts),
-         :ok <- conclude_task_runs(event, token, check_runs, results, opts) do
-      {:ok, results}
+  defp dispatch_from_source(event, token, source_path, janitor, opts) do
+    try do
+      maybe_after_source_acquired(source_path, opts)
+
+      with {:ok, graph, tasks} <- load_graph(source_path),
+           {:ok, check_runs} <- create_task_runs(event, token, tasks, opts),
+           :ok <- mark_in_progress(event, token, check_runs, opts),
+           {:ok, results} <- run_executor(tasks, graph, source_path, event.run_id, opts),
+           :ok <- conclude_task_runs(event, token, check_runs, results, opts) do
+        {:ok, results}
+      end
+    after
+      case workspace_janitor(opts).cleanup(janitor) do
+        :ok ->
+          :ok
+
+        {:error, :timeout} ->
+          Logger.warning("[GitHub Dispatcher] source workspace cleanup timed out")
+      end
     end
-  after
-    source_client(opts).cleanup(source_path, opts)
+  end
+
+  defp maybe_after_source_acquired(source_path, opts) do
+    case Keyword.get(opts, :after_source_acquired) do
+      callback when is_function(callback, 1) -> callback.(source_path)
+      _ -> :ok
+    end
   end
 
   defp create_suite(event, token, opts) do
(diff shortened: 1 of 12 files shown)


core/lib/sykli/github/dispatcher.ex
```diff
@@ -376,13 +378,8 @@ defmodule Sykli.GitHub.Dispatcher do
...
   defp source_client(opts), do: Keyword.get(opts, :source_client, Sykli.GitHub.Source)
 
-  defp retryable_dispatch_error?(%Sykli.Error{code: code}) do
-    code not in [
-      "github.app.missing_config",
-      "github.app.private_key_not_found",
-      "github.app.jwt_failed"
-    ]
-  end
+  defp retryable_dispatch_error?(%Sykli.Error{code: "github.app." <> _}), do: false
```
yairfalse: **The wildcard is too coarse for the current App taxonomy.**

Looking at `core/lib/sykli/github/app/real.ex`, the actual `github.app.*` codes today are:

| Code | Today's emit conditions | Permanent or transient? |
|---|---|---|
| `github.app.missing_config` | env var unset | Permanent |
| `github.app.private_key_not_found` | PEM file missing | Permanent |
| `github.app.jwt_failed` | `JOSE` signing crashed | Permanent (bad key) |
| `github.app.bad_response` | JSON decode / missing fields in token response | Mostly transient (GitHub blip), occasionally permanent |
| **`github.app.unauthorized`** | Two paths emit this — see below | **Both** |

The key issue is `github.app.unauthorized`. From `app/real.ex:84-103`:

```elixir
{:ok, code, body} ->        # non-2xx response from GitHub
  ...
  {:error, github_error("github.app.unauthorized", ..., {code, body})}

{:error, reason} ->          # transport-level failure (TCP RST, DNS, timeout)
  {:error, github_error("github.app.unauthorized", ..., reason)}
```

A TCP timeout to `api.github.com` (transient) emits the same code as a revoked installation (permanent). After this PR, both end up non-retryable — the timeout case will pin the delivery_id in the replay cache until the LRU TTL expires, blocking GitHub's automatic retry of a delivery that probably would have succeeded.

**Suggested follow-up** (out of scope for this PR — the storm-prevention win is real):

Split the overloaded code in `app/real.ex`:
- `github.app.unauthorized` — non-2xx HTTP response (permanent)
- `github.app.transport_failed` — `{:error, reason}` transport errors (transient)
- `github.app.upstream_error` — 5xx specifically (transient)

Then the dispatcher predicate stays a clean wildcard or moves to an explicit allow-list — either works once the codes themselves are honest.

**For this PR:** worth a `#` comment near this predicate noting that the wildcard is intentional and that the overloaded `github.app.unauthorized` taxonomy is a known follow-up. Otherwise the next reader will reasonably try to tighten this and not realize the dispatcher's coarseness is masking an upstream code conflation.

core/lib/sykli/github/dispatcher.ex
```diff
@@ -19,9 +19,11 @@ defmodule Sykli.GitHub.Dispatcher do
       {:error, %Sykli.Error{} = error} = result ->
         if retryable_dispatch_error?(error) do
           Deliveries.evict(delivery_id)
+          Logger.warning("[GitHub Dispatcher] dispatch failed", code: error.code)
+        else
+          Logger.warning("[GitHub Dispatcher] degraded mode - App unavailable", code: error.code)
```
yairfalse: **Two related observations on the new `else` branch:**

**1. Recovery UX:** once the operator fixes the App config (e.g., updates the private key), they'll typically click "Redeliver" in GitHub's webhook deliveries UI to retry the failed delivery. With this PR, that redelivery hits the replay cache (delivery_id still there because we didn't evict) → returns 409 → operator sees "Delivery failed" again with no obvious reason. They have to either wait for the LRU TTL to expire or restart the receiver. Worth a one-paragraph note in `docs/github-native.md` under "Recovery from App auth failures" — even just telling operators to wait N minutes or restart would close the loop.

**2. Log wording:** "degraded mode - App unavailable" reads like a system-level state, but this fires per-delivery. If the App is broken and 50 webhooks come in over an hour, you get 50 "degraded mode" warnings — looks alarming and might trigger separate paging/dashboards. Consider "App auth failed; delivery will not be retried" or similar — accurate per-event without implying system-wide state. The `code:` metadata already disambiguates the underlying cause for any operator who looks closely.

core/test/sykli/github/dispatcher_test.exs
```diff
@@ -107,4 +107,24 @@ defmodule Sykli.GitHub.DispatcherTest do
 
     assert {:error, :duplicate_delivery} = Deliveries.accept(event.delivery_id, 2)
   end
+
+  test "GitHub App authorization failures do not evict the delivery", %{event: event} do
```
yairfalse, line 114: **Coverage is sufficient for the wildcard but fragile against future tightening.**

This test covers `github.app.unauthorized` specifically. The current predicate is `"github.app." <> _`, so one example proves the rule. But if anyone in the future tightens the predicate back to an explicit list (which they might, given the overloaded `github.app.unauthorized` issue I flagged on dispatcher.ex:381), this single-code test won't catch the regression.

**Suggested:** add a second test for one of the obviously-permanent codes — `github.app.missing_config` is the cleanest:

```elixir
test "GitHub App config failures also do not evict the delivery", %{event: event} do
  assert :ok = Deliveries.accept(event.delivery_id, 1)

  assert {:error, %Sykli.Error{code: "github.app.missing_config"}} =
           Dispatcher.dispatch(event,
             app_client: Sykli.GitHub.App.Fake,
             app_response: {:error, %Sykli.Error{code: "github.app.missing_config", ...}}
           )

  assert {:error, :duplicate_delivery} = Deliveries.accept(event.delivery_id, 2)
end
```

Two tests across two distinct codes locks in "every github.app.* is non-retryable" as the contract. Future tightening of the predicate would have to either keep that contract or update both tests deliberately.

yairfalse (commented):
## Review summary

Focused fix for #152's retry storm. The shape is right: dispatch failures that GitHub will keep retrying must be classified as permanent so the replay cache can hold the delivery_id and short-circuit the retries with a 409. The PR widens the non-retryable predicate from an explicit 3-code list to a `github.app.*` wildcard and adds a regression test.

### Real concern

**The `github.app.*` wildcard is correct in spirit but currently lumps transient failures with permanent ones**, because `github.app.unauthorized` is overloaded in `app/real.ex:99-103` to cover both permanent (HTTP 401/403) AND transient (network errors via `{:error, reason}` from the HTTP client) failures. After this PR, a temporary network blip talking to `api.github.com` will mark the delivery non-retryable forever — the opposite of what the user wants. See inline on `dispatcher.ex:381`.

### Smaller findings

- Operator recovery UX after fixing config — replay cache still 409s until LRU expires. See inline on `dispatcher.ex:24`.
- "degraded mode" log wording fires per-delivery; misleading at scale. See inline on `dispatcher.ex:24`.
- Test covers one App code; a second case would harden against future regressions. See inline on `dispatcher_test.exs:111`.

### What's good

- Clear, minimal, scoped change
- Test added that matches the bug being fixed (delivery NOT evicted after App failure)
- Pattern-match on `"github.app." <> _` is idiomatic Elixir and obviously correct given a clean taxonomy
- Differentiated logging makes the operator's first response ("is this transient or do I need to look at config?") faster

### Recommendation

Approve. The overloaded `github.app.unauthorized` is a real correctness gap but it's an upstream taxonomy issue — fixable in a follow-up that splits the code in `app/real.ex`. This PR is still a net win against the storm scenario it's targeting.

core/lib/sykli/github/dispatcher.ex
yairfalse: Fixed in aa65774 by splitting the App error taxonomy instead of preserving the coarse wildcard. App 401/403 remains github.app.unauthorized and non-retryable; 5xx is now github.app.upstream_error and transport failures are github.app.transport_failed, both retryable. The dispatcher predicate is now an explicit permanent-code allowlist.

core/test/sykli/github/dispatcher_test.exs
yairfalse: Fixed in aa65774. Kept/merged the missing_config coverage from main, kept the unauthorized test, and added a transport_failed test proving retryable App transport failures still evict the delivery for GitHub retry.

core/lib/sykli/github/dispatcher.ex
yairfalse: Fixed in aa65774. Updated the per-delivery log line to "App auth failed; delivery will not be retried" and added a Recovery From App Auth Failures section to docs/github-native.md explaining the 409/redelivery behavior after config repair.

PR #160, opened 2026-05-03
fix(github): create task check runs in progress

### LINKED ISSUE #155: Dispatcher: mark_in_progress silently swallows transition errors, leaving Check Runs stuck at queued
**Severity:** Medium
**Found in:** PR #142 (merged) — `core/lib/sykli/github/dispatcher.ex:583-601`

## Problem

`mark_in_progress/4` calls `transition_check_run` for each task in `Enum.each`:

```elixir
defp mark_in_progress(event, token, check_runs, opts) do
  check_runs
  |> Enum.each(fn {task_name, check_run_id} ->
    transition_check_run(event, token, task_name, check_run_id,
                         "queued", "in_progress",
                         %{status: "in_progress"}, opts)
  end)

  :ok
end
```

`transition_check_run` only logs a warning on failure (and emits a `transition_failed` occurrence) — it never returns an error. `mark_in_progress` always returns `:ok` regardless of how many transitions actually succeeded.

Failure mode: GitHub rate-limits the burst of N concurrent PATCH calls (5000/hr per installation; a 30-task pipeline already burns ~60 calls). Every Check Run stays at `queued` until the task completes and the final completion update fires — which may itself rate-limit. Users see "queued" indefinitely with no signal that anything is wrong.

The `transition_failed` occurrence is broadcast but no consumer acts on it.

## Fix options

**Option A — abort on too many failures:**

```elixir
defp mark_in_progress(event, token, check_runs, opts) do
  failures = Enum.count(check_runs, fn {task_name, check_run_id} ->
    case transition_check_run(...) do
      :ok -> false
      {:error, _} -> true
    end
  end)

  cond do
    failures == 0 -> :ok
    failures >= map_size(check_runs) / 2 ->
      {:error, dispatch_error("github.dispatch.checks_unavailable",
                              "majority of in_progress transitions failed",
                              %{failed: failures, total: map_size(check_runs)})}
    true ->
      Logger.warning("[Dispatcher] some in_progress transitions failed",
                     failed: failures, total: map_size(check_runs))
      :ok
  end
end
```

**Option B — make `transition_check_run` actually return its result** so callers can decide, and let `mark_in_progress` be the policy point.

## Test that would catch this

```elixir
test "mark_in_progress aborts when majority of API calls fail" do
  # FakeChecks configured to return {:error, :rate_limited} for update_run.
  # Dispatcher should NOT proceed to executor; suite should conclude failure
  # with a github.dispatch.checks_unavailable error code.
end
```

### LINKED ISSUE #157: Dispatcher: 2 API calls per task to start (create-at-queued + update-to-in-progress)
**Severity:** Medium (efficiency / API budget)
**Found in:** PR #142 (merged) — `core/lib/sykli/github/dispatcher.ex`

## Problem

The dispatcher creates each Check Run at `queued`, then PATCHes it to `in_progress`, then PATCHes it again at `completed`:

- `create_task_runs/4` — POST `/repos/.../check-runs` (status: queued)
- `mark_in_progress/4` — PATCH `/repos/.../check-runs/<id>` (status: in_progress)
- `conclude_task_runs/5` — PATCH `/repos/.../check-runs/<id>` (status: completed)

That's 3 API calls per task. A 30-task pipeline = 90 calls, plus 1 suite-create = 91 calls per webhook.

GitHub Apps get 5000 req/hour per installation. That's a budget for ~55 webhooks per hour on a 30-task pipeline. Bursty CI loads (post-merge cascades, mass branch updates) can reach this in minutes.

The brief "queued" state has minimal UX value — users see it for milliseconds before tasks transition. The dispatcher could either:

1. **Create directly at `in_progress`.** Drop `mark_in_progress` entirely. POST with `status: "in_progress"` from the start. Saves 1 call per task (60 → 30 calls for 30 tasks).
2. **Create directly at the conclusion (skip both intermediate states).** Only viable if the executor result is known at create time, which it isn't (the create has to happen before the executor runs to set up the per-task UI).

Recommend option 1.

## Fix

```elixir
defp create_task_runs(event, token, tasks, opts) do
  Enum.reduce_while(tasks, {:ok, %{}}, fn task, {:ok, acc} ->
    create_opts = opts |> Keyword.put(:name, task.name) |> Keyword.put(:status, "in_progress")
    case checks_client(opts).create_run(...) do
      ...
```

And update `Sykli.GitHub.Checks.Real.create_run/3` to honor `opts[:status]` instead of hardcoding `"queued"`.

Then delete `mark_in_progress/4`.

## Test that would catch this

```elixir
test "create_run hits API with status=in_progress and skips the queued→in_progress transition" do
  # Dispatch a 5-task fixture.
  # Assert: 1 create_suite call, 5 create_run calls, 5 update_run calls (for completion).
  # Assert: zero update_run calls with %{status: "in_progress"}.
end
```

## Compatibility

The "queued" state was a Phase 1 implementation detail of the receiver creating one placeholder run; Phase 2 doesn't depend on it. ADR-021 §"Reporting" doesn't mandate a queued-then-in-progress transition.

yairfalse:
Summary:
- Creates per-task GitHub Check Runs directly with status in_progress.
- Removes the queued -> in_progress PATCH transition from the dispatcher.
- Keeps setup-failure check runs queued -> completed, since those are not task runs.
- Updates Checks.Real to honor opts[:status] and updates Fake/test assertions.

Closes #155.
Closes #157.

Validation:
- cd core && mix format
- cd core && env -u SYKLI_LABELS mix test test/sykli/github/dispatcher_test.exs test/sykli/github/checks_test.exs

### DIFF
--- a/core/lib/sykli/github/dispatcher.ex
+++ b/core/lib/sykli/github/dispatcher.ex
@@ -95,7 +95,6 @@ defmodule Sykli.GitHub.Dispatcher do
   defp dispatch_from_source(event, token, source_path, opts) do
     with {:ok, graph, tasks} <- load_graph(source_path),
          {:ok, check_runs} <- create_task_runs(event, token, tasks, opts),
-         :ok <- mark_in_progress(event, token, check_runs, opts),
          {:ok, results} <- run_executor(tasks, graph, source_path, event.run_id, opts),
          :ok <- conclude_task_runs(event, token, check_runs, results, opts) do
       {:ok, results}
@@ -167,7 +166,9 @@ defmodule Sykli.GitHub.Dispatcher do
       case checks_client(opts).create_run(
              %{repo: event.repo, head_sha: event.head_sha},
              token,
-             Keyword.put(opts, :name, task.name)
+             opts
+             |> Keyword.put(:name, task.name)
+             |> Keyword.put(:status, "in_progress")
            ) do
         {:ok, run} ->
           check_run_id = run["id"]
@@ -183,26 +184,6 @@ defmodule Sykli.GitHub.Dispatcher do
           {:halt, {:error, error}}
       end
     end)
-  end
-
-  defp mark_in_progress(event, token, check_runs, opts) do
-    check_runs
-    |> Enum.each(fn {task_name, check_run_id} ->
-      transition_check_run(
-        event,
-        token,
-        task_name,
-        check_run_id,
-        "queued",
-        "in_progress",
-        %{
-          status: "in_progress"
-        },
-        opts
-      )
-    end)
-
-    :ok
   end
 
   defp run_executor(tasks, graph, source_path, run_id, opts) do
--- a/core/test/sykli/github/dispatcher_test.exs
+++ b/core/test/sykli/github/dispatcher_test.exs
@@ -42,13 +42,13 @@ defmodule Sykli.GitHub.DispatcherTest do
                     "fake-installation-token-123"}
 
     assert_receive {:github_checks_create_run, %{head_sha: "abc123"},
-                    "fake-installation-token-123", "test"}
-
-    assert_receive {:github_checks_update_run, %{check_run_id: _}, "fake-installation-token-123",
-                    %{status: "in_progress"}}
+                    "fake-installation-token-123", "test", "in_progress"}
 
     assert_receive {:github_checks_update_run, %{check_run_id: _}, "fake-installation-token-123",
                     %{status: "completed", conclusion: "success"}}
+
+    refute_received {:github_checks_update_run, %{check_run_id: _}, "fake-installation-token-123",
+                     %{status: "in_progress"}}
 
     assert_receive %Sykli.Occurrence{type: "ci.github.run.dispatched"}
     assert_receive %Sykli.Occurrence{type: "ci.github.run.source_acquired"}
@@ -77,7 +77,7 @@ defmodule Sykli.GitHub.DispatcherTest do
              )
 
     assert_receive {:github_checks_create_run, %{head_sha: "abc123"},
-                    "fake-installation-token-123", "sykli/source"}
+                    "fake-installation-token-123", "sykli/source", "queued"}
 
     assert_receive {:github_checks_update_run, %{check_run_id: _}, "fake-installation-token-123",
                     %{status: "completed", conclusion: "failure"}}
--- a/core/lib/sykli/github/checks/fake.ex
+++ b/core/lib/sykli/github/checks/fake.ex
@@ -11,7 +11,11 @@ defmodule Sykli.GitHub.Checks.Fake do
 
   @impl true
   def create_run(context, token, opts \\ []) do
-    notify(opts, {:github_checks_create_run, context, token, Keyword.get(opts, :name)})
+    notify(
+      opts,
+      {:github_checks_create_run, context, token, Keyword.get(opts, :name),
+       Keyword.get(opts, :status, "queued")}
+    )
 
     case response(opts, :create_run_response, :default) do
       :default ->
--- a/core/lib/sykli/github/checks/real.ex
+++ b/core/lib/sykli/github/checks/real.ex
@@ -22,12 +22,13 @@ defmodule Sykli.GitHub.Checks.Real do
   @impl true
   def create_run(%{repo: repo, head_sha: head_sha}, token, opts \\ []) do
     name = Keyword.get(opts, :name, "sykli")
+    status = Keyword.get(opts, :status, "queued")
 
     request(
       :post,
       "/repos/#{repo}/check-runs",
       token,
-      %{name: name, head_sha: head_sha, status: "queued"},
+      %{name: name, head_sha: head_sha, status: status},
       opts,
       "github.checks.write_failed"
     )
--- a/core/test/sykli/github/checks_test.exs
+++ b/core/test/sykli/github/checks_test.exs
@@ -20,12 +20,13 @@ defmodule Sykli.GitHub.ChecksTest do
     assert Jason.decode!(body) == %{"head_sha" => "abc123"}
   end
 
-  test "create_run creates queued placeholder check run" do
+  test "create_run creates check run with requested status" do
     assert {:ok, %{"id" => 202}} =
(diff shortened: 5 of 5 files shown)


core/lib/sykli/github/dispatcher.ex
```diff
@@ -167,7 +166,9 @@ defmodule Sykli.GitHub.Dispatcher do
       case checks_client(opts).create_run(
              %{repo: event.repo, head_sha: event.head_sha},
              token,
-             Keyword.put(opts, :name, task.name)
+             opts
+             |> Keyword.put(:name, task.name)
+             |> Keyword.put(:status, "in_progress")
```
yairfalse, line 205: **FALSE Protocol observability change worth a CHANGELOG note.**

The removed `mark_in_progress` step used to fire a `ci.github.check_run.transitioned` occurrence per task with `from: "queued", to: "in_progress"`. After this PR, those events disappear — task check runs jump straight from `created` to the next `completed` transition.

I grepped the codebase: no internal subscribers, so nothing breaks here. But the FALSE Protocol is the agent-facing surface, and external consumers (Claude, Codex, downstream tools reading `.sykli/occurrence.json` or via MCP) may have been pattern-matching on those transitions for "task started" semantics.

The replacement signal is the existing `ci.github.check_run.created` occurrence — it already fires per task on line 176-179 of this file. Agents should treat "created" as "started" going forward.

**Suggested:** add a `Changed` entry to `CHANGELOG.md`:

> Per-task GitHub check runs are now created with `status: in_progress` directly. The transient `queued` state and the corresponding `ci.github.check_run.transitioned` (queued→in_progress) occurrences no longer fire for task runs. Setup-failure check runs (`sykli/source`) keep the queued→completed pattern unchanged.

Also worth referencing in the agent-side schema doc when it lands.

core/test/sykli/github/dispatcher_test.exs
```diff
@@ -42,14 +42,14 @@ defmodule Sykli.GitHub.DispatcherTest do
...
-                    "fake-installation-token-123", "test"}
-
-    assert_receive {:github_checks_update_run, %{check_run_id: _}, "fake-installation-token-123",
-                    %{status: "in_progress"}}
+                    "fake-installation-token-123", "test", "in_progress"}
 
     assert_receive {:github_checks_update_run, %{check_run_id: _}, "fake-installation-token-123",
                     %{status: "completed", conclusion: "success"}}
 
+    refute_received {:github_checks_update_run, %{check_run_id: _}, "fake-installation-token-123",
```
yairfalse, line 53: **`refute_received` is a current-mailbox-state check, not a never-arrives check.**

This assertion only catches the regression because of the *implicit* invariant that any `in_progress` update message would have arrived BEFORE the `completed` update message (the dispatcher sequences the API calls synchronously, so old-flow ordering would be `create → in_progress → completed`). By the time `assert_receive` for `completed` succeeds on line 47, any stray `in_progress` message would also be in the mailbox.

The assertion is correct, but a future reader could move the `refute_received` earlier in the test (before the create assertion) and silently break the regression check — `refute_received` would pass simply because no messages have arrived yet.

**Suggested:** add a one-line comment to make the ordering invariant explicit:

```elixir
# refute_received is a current-mailbox check; the assert_receive above
# guarantees the dispatcher has finished, so any in_progress update would
# already be queued.
refute_received {:github_checks_update_run, ...}
```

Not blocking; just protecting against a subtle future regression.

yairfalse (commented):
## Review summary

Clean simplification. Old flow: `create (queued) → update (in_progress) → run → update (completed)` — 3 API calls per task. New flow: `create (in_progress) → run → update (completed)` — 2 API calls per task. For a 50-task pipeline that's 50 fewer Checks API requests per dispatch — meaningful rate-limit headroom and a small latency win on "PR shows checks running."

The code is net-negative (32 deletions, 19 additions), the `mark_in_progress` helper is gone, and the GitHub Checks API supports `status: in_progress` directly on creation — so this is a real reduction, not a refactor that hides complexity elsewhere.

### Findings

1. **`ci.github.check_run.transitioned` events for queued→in_progress are gone.** No internal subscribers (verified via grep), but external agents reading FALSE Protocol occurrences may notice. Worth a CHANGELOG note. See inline on `dispatcher.ex:171`.
2. **`refute_received` is a "right now" check, not "ever."** Works here because of synchronous message ordering, but worth a comment. See inline on `dispatcher_test.exs:50`.

### Follow-up worth considering (out of scope)

**`create_setup_failure` (around `dispatcher.ex:255`, not in this diff) could apply the same simplification.** It still does the create-queued + transition-completed dance, which means 2 API calls for a state machine that only ever has one terminal value (`failure`). The GitHub Checks API allows `status: completed` + `conclusion: ...` + `output: {...}` all in the create request, so this could collapse to one call by extending `Checks.Real.create_run/3` to honor `:conclusion` and `:output` (it already honors `:status` per this PR). Same shape of redundant call this PR just eliminated for tasks — worth a follow-up issue.

### What's good

- Real perf / rate-limit win, not just code cleanup
- Backward-compatible default in `Checks.Real.create_run/3` (`status` defaults to `"queued"`)
- `Checks.Fake` notification message extended to include status — tests can now assert what status was requested
- `refute_received` regression assertion catches the "someone re-adds an in_progress update" anti-pattern
- `started_at` semantics improve incidentally — GitHub now records the actual start time at create, not the in_progress transition

### Recommendation

Approve. Add a CHANGELOG entry for the visible state-machine change. The setup_failure simplification can be a separate PR.

core/lib/sykli/github/dispatcher.ex
yairfalse: Fixed in 7111d0d. Added an Unreleased CHANGELOG Changed entry documenting that per-task GitHub check runs are now created as `in_progress`, and that the queued→in_progress `ci.github.check_run.transitioned` occurrence no longer fires for task runs while setup-failure runs keep the queued→completed path.

core/test/sykli/github/dispatcher_test.exs
yairfalse: Fixed in 7111d0d. Added a short comment above the `refute_received` explaining that the prior completed-update assertion proves synchronous Checks calls have finished, so any old queued→in_progress update would already be in the mailbox.

PR #161, opened 2026-05-03
fix(github): avoid walking source tree for size

### LINKED ISSUE #156: Dispatcher: directory_bytes walks entire cloned tree (including .git) for one occurrence field
**Severity:** Medium (perf)
**Found in:** PR #142 (merged) — `core/lib/sykli/github/dispatcher.ex:750-761`

## Problem

`directory_bytes/1` is called on every successful clone to populate the `bytes` field on a `ci.github.run.source_acquired` occurrence:

```elixir
defp directory_bytes(path) do
  path
  |> Path.join("**/*")
  |> Path.wildcard()
  |> Enum.filter(&File.regular?/1)
  |> Enum.reduce(0, fn file, acc ->
    case File.stat(file) do
      {:ok, stat} -> acc + stat.size
      {:error, _} -> acc
    end
  end)
end
```

For every webhook this walks the full cloned tree including `.git/objects` (which can be tens of thousands of files for a large repo, even with `--depth 1`). Each file triggers a `File.stat` syscall.

Real-world cost on a typical Elixir monorepo: 50–200 ms per webhook just for one informational field. On a 100-webhook-per-day mesh that's 10–20 wasted seconds of stat syscalls.

`Path.wildcard("**/*")` also loads the entire path list into memory before reducing — fine for small repos, problematic for very large ones.

## Fix options

1. **Drop the field.** Workspace size on disk is rarely actionable for the user; the head SHA + repo are already in the occurrence.
2. **Use `du`:**
   ```elixir
   defp directory_bytes(path) do
     case System.cmd("du", ["-sk", path], stderr_to_stdout: true) do
       {output, 0} ->
         output |> String.split() |> List.first() |> String.to_integer() |> Kernel.*(1024)
       _ -> nil
     end
   end
   ```
   One syscall, returns kilobytes. Set the field to `nil` on failure.
3. **Sample.** Compute size only for `lib/`, `src/`, top-level — cheaper proxy for "size of the user's code", excluding `.git`.

Recommend option 2 — same data, ~1000x cheaper.

## Test that would catch this

```elixir
@tag :performance
test "source acquisition stays under 100ms for a 10k-file fixture repo" do
  # Generate a fixture workspace with 10k empty files.
  # Time Source.acquire/3 + directory_bytes computation.
  # Assert: total elapsed < 100ms.
end
```

(Tag with `:performance`, not run by default.)

yairfalse:
Summary:
- Replaces the per-file Path.wildcard/File.stat source-size walk with a single du -sk call.
- Returns nil for bytes if du fails or emits unparsable output.
- Adds dispatcher coverage that verifies the du runner is used and bytes are converted from KiB to bytes.

Closes #156.

Validation:
- cd core && mix format

### DIFF
--- a/core/lib/sykli/github/dispatcher.ex
+++ b/core/lib/sykli/github/dispatcher.ex
@@ -131,7 +131,7 @@ defmodule Sykli.GitHub.Dispatcher do
           repo: event.repo,
           sha: event.head_sha,
           path: path,
-          bytes: directory_bytes(path)
+          bytes: directory_bytes(path, opts)
         })
 
         {:ok, path}
@@ -352,17 +352,28 @@ defmodule Sykli.GitHub.Dispatcher do
     end
   end
 
-  defp directory_bytes(path) do
-    path
-    |> Path.join("**/*")
-    |> Path.wildcard()
-    |> Enum.filter(&File.regular?/1)
-    |> Enum.reduce(0, fn file, acc ->
-      case File.stat(file) do
-        {:ok, stat} -> acc + stat.size
-        {:error, _} -> acc
-      end
-    end)
+  defp directory_bytes(path, opts) do
+    du_runner = Keyword.get(opts, :du_runner, &System.cmd/3)
+
+    case du_runner.("du", ["-sk", path], stderr_to_stdout: true) do
+      {output, 0} ->
+        output
+        |> String.split()
+        |> List.first()
+        |> parse_kibibytes()
+
+      _other ->
+        nil
+    end
+  end
+
+  defp parse_kibibytes(nil), do: nil
+
+  defp parse_kibibytes(value) do
+    case Integer.parse(value) do
+      {kib, ""} -> kib * 1024
+      _other -> nil
+    end
   end
 
   defp app_client(opts),
--- a/core/test/sykli/github/dispatcher_test.exs
+++ b/core/test/sykli/github/dispatcher_test.exs
@@ -51,9 +51,43 @@ defmodule Sykli.GitHub.DispatcherTest do
                     %{status: "completed", conclusion: "success"}}
 
     assert_receive %Sykli.Occurrence{type: "ci.github.run.dispatched"}
-    assert_receive %Sykli.Occurrence{type: "ci.github.run.source_acquired"}
+
+    assert_receive %Sykli.Occurrence{
+      type: "ci.github.run.source_acquired",
+      data: %{bytes: bytes}
+    }
+
+    assert is_integer(bytes)
     assert_receive %Sykli.Occurrence{type: "ci.github.check_run.created"}
     assert_receive %Sykli.Occurrence{type: "ci.github.check_suite.concluded"}
+  end
+
+  test "source_acquired bytes uses du instead of walking the tree", %{event: event} do
+    parent = self()
+
+    du_runner = fn "du", ["-sk", path], opts ->
+      send(parent, {:du_runner_called, path, opts})
+      {"7\t#{path}\n", 0}
+    end
+
+    assert :ok =
+             Dispatcher.dispatch(event,
+               app_client: Sykli.GitHub.App.Fake,
+               checks_client: Sykli.GitHub.Checks.Fake,
+               source_client: Sykli.GitHub.Source.Fake,
+               source_fixture: @fixture,
+               test_pid: self(),
+               fake_recorder: self(),
+               du_runner: du_runner
+             )
+
+    assert_receive {:du_runner_called, path, [stderr_to_stdout: true]}
+    assert String.ends_with?(path, "/repo")
+
+    assert_receive %Sykli.Occurrence{
+      type: "ci.github.run.source_acquired",
+      data: %{bytes: 7168}
+    }
   end
 
   test "dispatch failure evicts the delivery for GitHub retry", %{event: event} do


core/lib/sykli/github/dispatcher.ex
```diff
@@ -352,17 +352,28 @@ defmodule Sykli.GitHub.Dispatcher do
...
-    |> Enum.reduce(0, fn file, acc ->
-      case File.stat(file) do
-        {:ok, stat} -> acc + stat.size
-        {:error, _} -> acc
-      end
-    end)
+  defp directory_bytes(path, opts) do
+    du_runner = Keyword.get(opts, :du_runner, &System.cmd/3)
+
+    case du_runner.("du", ["-sk", path], stderr_to_stdout: true) do
```
yairfalse, line 380: **Output parsing is brittle to mixed stderr + stdout.**

`stderr_to_stdout: true` merges `du`'s stderr into the captured output. If `du` traverses a directory it can't read (permission denied, broken symlink, race with cleanup), it emits a warning like:

```
du: cannot read directory '/path/to/foo': Permission denied
42\t/path/to/source
```

With `String.split()` (whitespace-split), the resulting tokens are `["du:", "cannot", "read", ..., "42", "/path/to/source"]`. `List.first()` returns `"du:"` → `Integer.parse/1` returns `:error` → `parse_kibibytes/1` returns `nil`.

The `du` exit code is still `0` in this case (warnings don't fail the command), so the `{output, 0}` branch matches and we silently drop a size we could have parsed.

**Suggested fix:** parse the *last line* (the summary line is always last), then take the *first token* of that line:

```elixir
case du_runner.("du", ["-sk", path], stderr_to_stdout: true) do
  {output, 0} ->
    output
    |> String.split("\n", trim: true)
    |> List.last()
    |> case do
      nil -> nil
      line -> line |> String.split() |> List.first() |> parse_kibibytes()
    end

  _other ->
    nil
end
```

Or drop `stderr_to_stdout: true` so warnings stay separate from the summary.

core/lib/sykli/github/dispatcher.ex
```diff
@@ -352,17 +352,28 @@ defmodule Sykli.GitHub.Dispatcher do
...
-    |> Path.join("**/*")
-    |> Path.wildcard()
-    |> Enum.filter(&File.regular?/1)
-    |> Enum.reduce(0, fn file, acc ->
-      case File.stat(file) do
-        {:ok, stat} -> acc + stat.size
-        {:error, _} -> acc
-      end
-    end)
+  defp directory_bytes(path, opts) do
```
yairfalse, line 377: **Behavior change in `bytes` semantics — worth a CHANGELOG entry.**

Two subtle shifts vs the old `Path.wildcard` + `File.stat` walk:

1. **Disk usage vs apparent size.** `du -sk` reports disk usage (block count × block size). The old code summed `stat.size` (apparent size). For a tree of many small files (e.g., a Node.js node_modules), `du` can be 2–10× larger than apparent size due to per-file block rounding. For a repo with a few large files, they'll be roughly equal.
2. **Hidden files now included.** `Path.wildcard("**/*")` defaults to skipping dotfiles, so `.git/` was *excluded* from the old measurement. `du` includes everything. Since `Source.Real` does `git clone --depth 1`, the `.git/` directory is real and can be a meaningful chunk of the total — so the reported size will jump.

Neither is wrong; both are arguably more honest. But anyone watching the `bytes` field in occurrences will see a step change. Worth a `Changed` entry in CHANGELOG.

If you wanted strict equivalence with the old behavior, `du -sk --apparent-size --exclude=.git` (GNU-only) would do it — but the cross-platform/portability cost isn't worth the fidelity here.

core/lib/sykli/github/dispatcher.ex
```diff
@@ -352,17 +352,28 @@ defmodule Sykli.GitHub.Dispatcher do
...
+
+    case du_runner.("du", ["-sk", path], stderr_to_stdout: true) do
+      {output, 0} ->
+        output
+        |> String.split()
+        |> List.first()
+        |> parse_kibibytes()
+
+      _other ->
+        nil
```
yairfalse, line 388: **`nil` vs `0` on failure — downstream consumer impact.**

The old `directory_bytes/1` always returned an integer (0 if every stat failed, but never nil). The new code returns `nil` when `du` fails or output is unparseable.

The semantic improvement is real ("we don't know" beats "we computed and got zero"), but any agent or downstream consumer doing arithmetic on the `bytes` field will now hit `nil` and crash unless they handle it. The existing test on line 60 of dispatcher_test.exs already asserts `is_integer(bytes)` — that assertion will fail on the new failure path.

Not blocking — this is the right change. But:
- Worth a `Changed` note in CHANGELOG documenting the shape shift.
- Worth a quick audit of any consumer that reads `data.bytes` from `ci.github.run.source_acquired` (probably just `.sykli/` storage and any agent consuming it via MCP/query).

core/test/sykli/github/dispatcher_test.exs
```diff
@@ -51,11 +51,45 @@ defmodule Sykli.GitHub.DispatcherTest do
...
+      type: "ci.github.run.source_acquired",
+      data: %{bytes: bytes}
+    }
+
+    assert is_integer(bytes)
     assert_receive %Sykli.Occurrence{type: "ci.github.check_run.created"}
     assert_receive %Sykli.Occurrence{type: "ci.github.check_suite.concluded"}
   end
 
+  test "source_acquired bytes uses du instead of walking the tree", %{event: event} do
```
yairfalse, line 70: **Test only covers the happy path.** Worth adding cases for:

- **Non-zero exit:** `du_runner` returns `{"du: ...", 1}` → `bytes: nil`
- **Unparseable output:** `du_runner` returns `{"garbage\n", 0}` → `bytes: nil`
- **Empty output:** `du_runner` returns `{"", 0}` → `bytes: nil`
- **Mixed warning + summary** (the parsing brittleness from the inline above): `du_runner` returns `{"du: cannot read ...\n42\t/path\n", 0}` — currently broken, would catch the parsing bug

These are cheap to add (each is ~10 lines, same shape as the happy-path test) and they lock in the failure semantics. The mixed-warning case is the one most likely to fire in production — webhook receivers run with whatever permissions they have, and PRs occasionally include weird files that `du` complains about.

yairfalse (commented):
## Review summary

Replaces a per-file `Path.wildcard` + `File.stat` walk with a single `du -sk` call for source-workspace size measurement. Closes #156.

The performance win is real — for a large repo (think Linux kernel scale, ~80k files), the old code does ~80k `lstat` syscalls plus a directory traversal in BEAM's IO scheduler; `du -sk` is one fork+exec to a native C implementation that uses optimized traversal. Even for sykli-sized repos, this is one syscall vs hundreds.

### Real concerns

1. **Output parsing is brittle to mixed stderr + stdout.** `stderr_to_stdout: true` plus `List.first()` on `String.split()` will misparse if `du` emits warnings before the summary. See inline on `dispatcher.ex:358`.
2. **Behavior change in `bytes` semantics** — disk usage with block-size rounding, plus now including hidden files like `.git/`. Worth a CHANGELOG entry. See inline on `dispatcher.ex:355`.
3. **`nil` vs `0` on failure** — semantic shift downstream consumers may not expect. See inline on `dispatcher.ex:366`.
4. **No tests for failure modes** — see inline on `dispatcher_test.exs:65`.

### What's good

- Major performance improvement for any non-trivial workspace size
- `:du_runner` injection pattern matches existing `:source_client` / `:checks_client` conventions
- `nil` return on failure is semantically more honest than the old `0`-on-stat-failure (which silently masked errors)
- Test asserts the runner is called with the right args + the KiB→bytes conversion (good happy-path coverage)
- `parse_kibibytes/1` extracted as a small, well-named helper

### Recommendation

Approve with concern 1 fixed in this PR (real bug). Concerns 2–4 can land separately or be addressed alongside.

core/lib/sykli/github/dispatcher.ex
yairfalse: Fixed in 95ebee5. The parser now reads the last non-empty `du` output line before parsing the first token, so warning lines merged through stderr no longer hide a valid summary.

core/lib/sykli/github/dispatcher.ex
yairfalse: Fixed in 95ebee5. Added a CHANGELOG Changed entry documenting the `du -sk` semantics: disk usage instead of apparent size, dotfiles included, and the resulting step change in `ci.github.run.source_acquired.data.bytes`.

core/lib/sykli/github/dispatcher.ex
yairfalse: Fixed in 95ebee5. The same CHANGELOG entry now documents that an uncomputable source-workspace size is represented as `bytes: null` instead of `0`. I also audited in-repo consumers with `rg source_acquired|bytes`; only occurrence production, PubSub, docs, and tests reference this field.

core/test/sykli/github/dispatcher_test.exs
yairfalse: Fixed in 95ebee5. Added tests for mixed warning+summary output, non-zero `du`, unparseable output, and empty output. These lock in both the parser behavior and the `bytes: nil` failure semantics.

PR #162, opened 2026-05-03
fix(github): clean dispatcher workspace after crashes

yairfalse:
## Summary
- add a monitor-backed workspace janitor for GitHub dispatcher source directories
- keep normal-path cleanup synchronous while preserving cleanup after dispatcher kill/crash
- add a regression test that kills the dispatcher after source acquisition and asserts the workspace is removed

Closes #149

## Validation
- cd core && mix format
- cd core && mix credo
- cd core && env -u SYKLI_LABELS mix test

### DIFF
--- a/core/lib/sykli/github/dispatcher.ex
+++ b/core/lib/sykli/github/dispatcher.ex
@@ -93,15 +93,43 @@ defmodule Sykli.GitHub.Dispatcher do
   end
 
   defp dispatch_from_source(event, token, source_path, opts) do
-    with {:ok, graph, tasks} <- load_graph(source_path),
-         {:ok, check_runs} <- create_task_runs(event, token, tasks, opts),
-         :ok <- mark_in_progress(event, token, check_runs, opts),
-         {:ok, results} <- run_executor(tasks, graph, source_path, event.run_id, opts),
-         :ok <- conclude_task_runs(event, token, check_runs, results, opts) do
-      {:ok, results}
+    with_workspace_janitor(source_path, opts, fn ->
+      maybe_after_source_acquired(source_path, opts)
+
+      with {:ok, graph, tasks} <- load_graph(source_path),
+           {:ok, check_runs} <- create_task_runs(event, token, tasks, opts),
+           :ok <- mark_in_progress(event, token, check_runs, opts),
+           {:ok, results} <- run_executor(tasks, graph, source_path, event.run_id, opts),
+           :ok <- conclude_task_runs(event, token, check_runs, results, opts) do
+        {:ok, results}
+      end
+    end)
+  end
+
+  defp with_workspace_janitor(source_path, opts, fun) do
+    case workspace_janitor(opts).start(self(), source_path, opts) do
+      {:ok, janitor} ->
+        try do
+          fun.()
+        after
+          workspace_janitor(opts).cleanup(janitor)
+        end
+
+      {:error, reason} ->
+        {:error,
+         dispatch_error(
+           "github.dispatch.workspace_janitor_failed",
+           "failed to monitor source workspace cleanup",
+           reason
+         )}
     end
-  after
-    source_client(opts).cleanup(source_path, opts)
+  end
+
+  defp maybe_after_source_acquired(source_path, opts) do
+    case Keyword.get(opts, :after_source_acquired) do
+      callback when is_function(callback, 1) -> callback.(source_path)
+      _ -> :ok
+    end
   end
 
   defp create_suite(event, token, opts) do
@@ -375,6 +403,9 @@ defmodule Sykli.GitHub.Dispatcher do
 
   defp checks_client(opts), do: Keyword.get(opts, :checks_client, Sykli.GitHub.Checks)
   defp source_client(opts), do: Keyword.get(opts, :source_client, Sykli.GitHub.Source)
+
+  defp workspace_janitor(opts),
+    do: Keyword.get(opts, :workspace_janitor, Sykli.GitHub.WorkspaceJanitor)
 
   defp retryable_dispatch_error?(%Sykli.Error{code: code}) do
     code not in [
--- a/core/lib/sykli/github/workspace_janitor.ex
+++ b/core/lib/sykli/github/workspace_janitor.ex
@@ -1,0 +1,51 @@
+defmodule Sykli.GitHub.WorkspaceJanitor do
+  @moduledoc "Cleans GitHub source workspaces when their owner process exits."
+
+  require Logger
+
+  @cleanup_timeout_ms 5_000
+
+  @spec start(pid(), String.t(), keyword()) :: {:ok, pid()} | {:error, term()}
+  def start(owner, path, opts \\ []) when is_pid(owner) and is_binary(path) do
+    Task.start(fn ->
+      ref = Process.monitor(owner)
+      loop(owner, ref, path, opts)
+    end)
+  end
+
+  @spec cleanup(pid()) :: :ok
+  def cleanup(pid) when is_pid(pid) do
+    ref = make_ref()
+    send(pid, {:cleanup, self(), ref})
+
+    receive do
+      {^ref, :ok} -> :ok
+    after
+      @cleanup_timeout_ms -> :ok
+    end
+  end
+
+  def cleanup(_pid), do: :ok
+
+  defp loop(owner, monitor_ref, path, opts) do
+    receive do
+      {:cleanup, caller, reply_ref} ->
+        Process.demonitor(monitor_ref, [:flush])
+        do_cleanup(path, opts)
+        send(caller, {reply_ref, :ok})
+
+      {:DOWN, ^monitor_ref, :process, ^owner, _reason} ->
+        do_cleanup(path, opts)
+    end
+  end
+
+  defp do_cleanup(path, opts) do
+    source_client(opts).cleanup(path, opts)
+  rescue
+    error ->
+      Logger.warning("[GitHub WorkspaceJanitor] cleanup failed", error: inspect(error))
+      :ok
+  end
+
+  defp source_client(opts), do: Keyword.get(opts, :source_client, Sykli.GitHub.Source)
+end
--- a/core/test/sykli/github/dispatcher_test.exs
+++ b/core/test/sykli/github/dispatcher_test.exs
@@ -107,4 +107,51 @@ defmodule Sykli.GitHub.DispatcherTest do
 
     assert {:error, :duplicate_delivery} = Deliveries.accept(event.delivery_id, 2)
   end
+
+  test "dispatch cleans up the source workspace when the dispatcher process is killed", %{
+    event: event
+  } do
+    parent = self()
+    event = %{event | delivery_id: "dispatcher-crash-cleanup"}
+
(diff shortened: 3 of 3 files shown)


yairfalse (commented):
## Review summary

Focused fix for #149 with the right shape: the OTP-idiomatic "monitor + janitor" pattern is exactly how you handle "cleanup must run even if owner is `kill`-ed" — `try/after` doesn't help there because `kill` skips Elixir's exception machinery entirely. The janitor is correctly unsupervised (must outlive the dispatcher), monitors the owner from inside the spawned task (which handles the dead-PID race via `Process.monitor`'s immediate-`:DOWN` semantics), and demonitors before the synchronous cleanup path to prevent double-cleanup. Good design.

### Real concerns

1. **Race window between `acquire_source` and `dispatch_from_source`** — the janitor doesn't exist yet during this gap. See inline on `dispatcher.ex:55`.
2. **`cleanup/1` silently returns `:ok` on timeout** — the dispatcher proceeds thinking cleanup happened. See inline on `workspace_janitor.ex:24`.
3. **The `Task.start` error branch leaks the source** — see inline on `dispatcher.ex:118`.

### Smaller findings

- `rescue` doesn't catch Erlang exits (`workspace_janitor.ex:44`)
- `after_source_acquired` test hook in production code (`dispatcher.ex:128`)
- Test only covers the crash path; the normal-exit path is implicit (`dispatcher_test.exs:111`)

### What's good

- Janitor is correctly unsupervised — must outlive the dispatcher to clean up after a crash
- `Process.demonitor(ref, [:flush])` before synchronous cleanup correctly prevents double-execution if `:DOWN` arrives mid-cleanup
- The injection pattern (`workspace_janitor(opts)`) matches existing `checks_client(opts)` / `source_client(opts)` conventions
- The regression test directly exercises the bug (kill the dispatcher, assert workspace gone) — exactly the right shape
- `assert_eventually` with `rescue ExUnit.AssertionError` is the standard ExUnit polling pattern

### Recommendation

Approve with concerns 1–3 addressed. Smaller findings can land separately or be left.

core/lib/sykli/github/dispatcher.ex
```diff
@@ -93,15 +93,43 @@ defmodule Sykli.GitHub.Dispatcher do
   end
 
   defp dispatch_from_source(event, token, source_path, opts) do
```
yairfalse: **Race window upstream of this function.** The janitor is created here in `dispatch_from_source`, but `acquire_source` was already called on the previous line of `dispatch_after_suite` (line 55, not in this diff):

```elixir
with {:ok, source_path} <- acquire_source(event, token, opts),
     {:ok, results} <- dispatch_from_source(event, token, source_path, opts) do
```

If the dispatcher process is killed *between* `acquire_source` returning and `dispatch_from_source` entering, the workspace is acquired but no janitor exists yet — same leak this PR otherwise fixes.

The window is tiny and #149's main scenario (crash during executor run) is fixed. But for completeness, consider moving the janitor setup *into* `acquire_source` itself (start the janitor immediately after the source is acquired, before returning the path) so the only remaining race is internal to `Source.Real.acquire` — much harder to lose to.

Acceptable as a follow-up if you'd rather keep this PR focused; worth at least a `# TODO` comment so the gap doesn't get forgotten.

core/lib/sykli/github/workspace_janitor.ex
```diff
@@ -0,0 +1,51 @@
...
+
+  @spec cleanup(pid()) :: :ok
+  def cleanup(pid) when is_pid(pid) do
+    ref = make_ref()
+    send(pid, {:cleanup, self(), ref})
+
+    receive do
+      {^ref, :ok} -> :ok
+    after
+      @cleanup_timeout_ms -> :ok
```
yairfalse: **Silent timeout swallows a real failure mode.** If `do_cleanup` takes >5s (large workspace, slow IO, hung NFS, locked file on Windows-style FS, etc.), `cleanup/1` returns `:ok` and the dispatcher proceeds as if cleanup completed. The janitor process meanwhile keeps running and eventually finishes cleanup (or gets stuck) — but the caller has no idea.

**Suggested fix:** at minimum log a warning on the timeout branch. Better, return `{:error, :timeout}` and have the dispatcher emit an occurrence so it shows up in observability. The `@spec` would need to widen but that's the right shape.

```elixir
after
  @cleanup_timeout_ms ->
    Logger.warning("[GitHub WorkspaceJanitor] cleanup timeout after #{@cleanup_timeout_ms}ms", path: ...)
    {:error, :timeout}
end
```

core/lib/sykli/github/dispatcher.ex
```diff
@@ -93,15 +93,43 @@ defmodule Sykli.GitHub.Dispatcher do
...
+  defp with_workspace_janitor(source_path, opts, fun) do
+    case workspace_janitor(opts).start(self(), source_path, opts) do
+      {:ok, janitor} ->
+        try do
+          fun.()
+        after
+          workspace_janitor(opts).cleanup(janitor)
+        end
+
+      {:error, reason} ->
```
yairfalse: **This error branch leaks the source workspace.** If `workspace_janitor.start/3` returns `{:error, ...}`, `dispatch_from_source` returns the error without ever cleaning up `source_path` — which was already acquired before this function was called.

In practice `Task.start` essentially never fails (only on `:erlang.system_info(:process_limit)` exhaustion), so this branch is dead in normal operation. But if it ever fires, the leak is a regression vs. the old `after` cleanup behavior.

**Suggested fix:** do best-effort cleanup in this branch:

```elixir
{:error, reason} ->
  source_client(opts).cleanup(source_path, opts)
  {:error, dispatch_error("github.dispatch.workspace_janitor_failed", ...)}
```

Or drop the error branch entirely and `Task.start!`-equivalent — the failure mode is so rare that crashing the dispatcher is also defensible.

core/lib/sykli/github/workspace_janitor.ex
```diff
@@ -0,0 +1,51 @@
...
+        send(caller, {reply_ref, :ok})
+
+      {:DOWN, ^monitor_ref, :process, ^owner, _reason} ->
+        do_cleanup(path, opts)
+    end
+  end
+
+  defp do_cleanup(path, opts) do
+    source_client(opts).cleanup(path, opts)
+  rescue
```
yairfalse, line 50: **`rescue` only catches Elixir exceptions, not Erlang exits.** If `source_client.cleanup/2` calls a GenServer that times out, or otherwise exits (`exit/1`), the janitor crashes silently — the `rescue` clause doesn't fire. For a best-effort cleanup task that's probably fine (the OS will GC the temp dir eventually), but worth being explicit about.

If you want to catch exits too:

```elixir
defp do_cleanup(path, opts) do
  source_client(opts).cleanup(path, opts)
rescue
  error -> Logger.warning(...); :ok
catch
  :exit, reason -> Logger.warning("[...] cleanup exited", reason: inspect(reason)); :ok
end
```

Not blocking; just flagging that the current rescue is narrower than it might appear.

core/lib/sykli/github/dispatcher.ex
```diff
@@ -93,15 +93,43 @@ defmodule Sykli.GitHub.Dispatcher do
...
+        {:error,
+         dispatch_error(
+           "github.dispatch.workspace_janitor_failed",
+           "failed to monitor source workspace cleanup",
+           reason
+         )}
+    end
+  end
+
+  defp maybe_after_source_acquired(source_path, opts) do
```
yairfalse, line 117: **Test seam in production code.** `after_source_acquired` exists only so the regression test can synchronize on "source is acquired, janitor is alive, kill now." It's a pragmatic choice — the alternative (subscribing to a `:telemetry` event the test fires on) is heavier and the project doesn't seem to use telemetry for internal lifecycle yet.

Worth flagging because production code now carries a hook that only tests use. If you ever add a few more of these, consider a unified `:telemetry`-based hook system — at one or two it's not worth the abstraction. No change needed for this PR.

core/test/sykli/github/dispatcher_test.exs
```diff
@@ -107,4 +107,51 @@ defmodule Sykli.GitHub.DispatcherTest do
 
     assert {:error, :duplicate_delivery} = Deliveries.accept(event.delivery_id, 2)
   end
+
+  test "dispatch cleans up the source workspace when the dispatcher process is killed", %{
```
yairfalse, line 114: **Test only covers the crash path.** The other path through the janitor — `cleanup/1` being called explicitly when `dispatch_from_source` exits normally, triggering the `{:cleanup, caller, reply_ref}` branch in `loop/4` — isn't directly tested by this new case. It's almost certainly exercised by the existing dispatch tests (which presumably finish without crashing), but the assertion that the workspace is *also* cleaned up on normal exit is implicit.

Would be worth either:
- A direct unit test for `Sykli.GitHub.WorkspaceJanitor` covering both branches of `loop/4` in isolation (cheap, ~20 LOC).
- Or a quick assertion in an existing happy-path dispatch test that `source_path` doesn't exist after dispatch returns.

The second option also catches a different bug: if cleanup ever stops being called on the normal path (refactor regression), this test would catch it.

core/lib/sykli/github/dispatcher.ex
yairfalse: Fixed in 9470b9f. The janitor is now started inside acquire_source/3 immediately after source acquisition and before the path is returned to dispatch_after_suite/4, so the previous acquire_source -> dispatch_from_source race window is closed.

core/lib/sykli/github/workspace_janitor.ex
yairfalse: Fixed in 9470b9f. WorkspaceJanitor.cleanup/1 now logs a timeout warning and returns {:error, :timeout}; the dispatcher logs when synchronous source cleanup times out instead of silently treating it as complete.

core/lib/sykli/github/dispatcher.ex
yairfalse: Fixed in 9470b9f. If janitor startup fails after source acquisition, acquire_source/3 now performs best-effort Source.cleanup/2 before returning github.dispatch.workspace_janitor_failed. Added a regression test for this branch.

core/lib/sykli/github/workspace_janitor.ex
yairfalse: Handled in 9470b9f. do_cleanup/2 now catches exits as well as exceptions and logs cleanup exited with the inspected exit reason.

core/lib/sykli/github/dispatcher.ex
yairfalse: Left this as-is for now. The hook remains a narrow lifecycle seam used only to synchronize the crash regression; I agree that if more of these appear, we should replace them with a proper telemetry/lifecycle hook rather than ad hoc opts.

core/test/sykli/github/dispatcher_test.exs
yairfalse: Fixed in 9470b9f. The happy-path dispatcher test now asserts it receives github_source_cleanup and that the workspace path no longer exists after dispatch returns, so the explicit cleanup branch is covered.

PR #163, opened 2026-05-03
fix(github): clean up dispatcher nits

yairfalse:
## Summary
- unify GitHub source injection through the Source facade using the existing :impl option
- make source workspace path segments avoid ':' for better portability
- derive suite occurrence conclusions from CheckRunFormatter's per-task conclusion mapping
- strengthen source cleanup path-containment test
- replace fragile receiver retry polling with an explicit dispatch-completed signal in the test double

Closes #158

## Validation
- cd core && mix format
- cd core && env -u SYKLI_LABELS mix test test/sykli/github/dispatcher_test.exs test/sykli/github/source_test.exs test/sykli/github/webhook/receiver_test.exs
- cd core && mix credo

Full suite note:
- cd core && env -u SYKLI_LABELS mix test hit the existing StreamData TooManyDuplicatesError flake in Sykli.Mesh.Transport.Sim.EventQueueTest; no other failures were reported.

### DIFF
--- a/core/lib/sykli/github/dispatcher.ex
+++ b/core/lib/sykli/github/dispatcher.ex
@@ -58,7 +58,7 @@ defmodule Sykli.GitHub.Dispatcher do
         repo: event.repo,
         head_sha: event.head_sha,
         check_suite_id: suite["id"],
-        conclusion: conclusion(results)
+        conclusion: suite_conclusion(results)
       })
 
       :ok
@@ -330,7 +330,11 @@ defmodule Sykli.GitHub.Dispatcher do
     end
   end
 
-  defp conclusion(results) do
+  @doc false
+  @spec suite_conclusion([Sykli.Executor.TaskResult.t()]) :: String.t()
+  def suite_conclusion([]), do: "success"
+
+  def suite_conclusion(results) do
     conclusions = Enum.map(results, &CheckRunFormatter.conclusion/1)
 
     cond do
--- a/core/lib/sykli/github/dispatcher.ex
+++ b/core/lib/sykli/github/dispatcher.ex
@@ -101,7 +101,7 @@ defmodule Sykli.GitHub.Dispatcher do
       {:ok, results}
     end
   after
-    source_client(opts).cleanup(source_path, opts)
+    Sykli.GitHub.Source.cleanup(source_path, opts)
   end
 
   defp create_suite(event, token, opts) do
@@ -125,7 +125,7 @@ defmodule Sykli.GitHub.Dispatcher do
   end
 
   defp acquire_source(event, token, opts) do
-    case source_client(opts).acquire(event, token, opts) do
+    case Sykli.GitHub.Source.acquire(event, token, opts) do
       {:ok, path} ->
         OccPubSub.github_run_source_acquired(event.run_id, %{
           repo: event.repo,
@@ -331,10 +331,13 @@ defmodule Sykli.GitHub.Dispatcher do
   end
 
   defp conclusion(results) do
-    if Enum.any?(results, &(&1.status in [:failed, :errored, :blocked])) do
-      "failure"
-    else
-      "success"
+    conclusions = Enum.map(results, &CheckRunFormatter.conclusion/1)
+
+    cond do
+      Enum.any?(conclusions, &(&1 == "failure")) -> "failure"
+      Enum.any?(conclusions, &(&1 == "cancelled")) -> "cancelled"
+      Enum.all?(conclusions, &(&1 == "skipped")) -> "skipped"
+      true -> "success"
     end
   end
 
@@ -374,7 +377,6 @@ defmodule Sykli.GitHub.Dispatcher do
       )
 
   defp checks_client(opts), do: Keyword.get(opts, :checks_client, Sykli.GitHub.Checks)
-  defp source_client(opts), do: Keyword.get(opts, :source_client, Sykli.GitHub.Source)
 
   defp retryable_dispatch_error?(%Sykli.Error{code: code}) do
     code not in [
--- a/core/test/sykli/github/dispatcher_test.exs
+++ b/core/test/sykli/github/dispatcher_test.exs
@@ -32,7 +32,7 @@ defmodule Sykli.GitHub.DispatcherTest do
              Dispatcher.dispatch(event,
                app_client: Sykli.GitHub.App.Fake,
                checks_client: Sykli.GitHub.Checks.Fake,
-               source_client: Sykli.GitHub.Source.Fake,
+               impl: Sykli.GitHub.Source.Fake,
                source_fixture: @fixture,
                test_pid: self(),
                fake_recorder: self()
@@ -63,7 +63,7 @@ defmodule Sykli.GitHub.DispatcherTest do
              Dispatcher.dispatch(event,
                app_client: Sykli.GitHub.App.Fake,
                checks_client: Sykli.GitHub.Checks.Fake,
-               source_client: Sykli.GitHub.Source.Fake,
+               impl: Sykli.GitHub.Source.Fake,
                test_pid: self(),
                source_response:
                  {:error,
--- a/core/test/sykli/github/dispatcher_test.exs
+++ b/core/test/sykli/github/dispatcher_test.exs
@@ -4,6 +4,7 @@ defmodule Sykli.GitHub.DispatcherTest do
   alias Sykli.GitHub.Dispatcher
   alias Sykli.GitHub.Webhook.Deliveries
   alias Sykli.Occurrence.PubSub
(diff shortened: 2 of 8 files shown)


core/lib/sykli/github/dispatcher.ex
```diff
@@ -331,10 +331,13 @@ defmodule Sykli.GitHub.Dispatcher do
   end
 
   defp conclusion(results) do
```
yairfalse: **Behavior change, not just a refactor — worth a CHANGELOG entry.**

The old mapping was:
- `any failed/errored/blocked` → `"failure"`
- else → `"success"`

The new mapping (via `CheckRunFormatter.conclusion/1`) is:
- `any failed/errored` → `"failure"`
- else `any blocked` → `"cancelled"` *(new — was `"failure"`)*
- else `all skipped` → `"skipped"` *(new — was `"success"`)*
- else → `"success"`

User-visible at the GitHub Check Suite level: a suite with only `:blocked` tasks now reports `cancelled` instead of `failure`; a suite with only `:skipped` tasks now reports `skipped` instead of `success`.

These are arguably more correct (a `:blocked` task didn't fail, it was prevented from running by an upstream failure; an all-skipped suite isn't a success), but they're changes downstream consumers (PR badges, branch protection rules, anyone scripting against check suite conclusions) will see.

**Suggested:** add a `Changed` entry in `CHANGELOG.md` under the unreleased / 0.6.x section.

core/lib/sykli/github/dispatcher.ex
```diff
@@ -331,10 +331,13 @@ defmodule Sykli.GitHub.Dispatcher do
...
-    if Enum.any?(results, &(&1.status in [:failed, :errored, :blocked])) do
-      "failure"
-    else
-      "success"
+    conclusions = Enum.map(results, &CheckRunFormatter.conclusion/1)
+
+    cond do
+      Enum.any?(conclusions, &(&1 == "failure")) -> "failure"
+      Enum.any?(conclusions, &(&1 == "cancelled")) -> "cancelled"
+      Enum.all?(conclusions, &(&1 == "skipped")) -> "skipped"
```
yairfalse, line 343: **Empty results map to `"skipped"` via vacuous truth.**

`Enum.all?([], _)` returns `true` in Elixir, so `conclusion([])` returns `"skipped"`. That's almost certainly wrong — an empty results list means no tasks ran (degenerate pipeline), which is closer to `"success"` or `"neutral"` than `"skipped"`.

This is unlikely to fire in practice (`load_graph` would presumably reject an empty pipeline upstream), but it's the kind of edge case that surfaces as "why did my empty pipeline report skipped?" months later.

**Suggested fix:** guard the empty case explicitly:

```elixir
defp conclusion([]), do: "success"
defp conclusion(results) do
  conclusions = Enum.map(results, &CheckRunFormatter.conclusion/1)
  cond do
    Enum.any?(conclusions, &(&1 == "failure")) -> "failure"
    Enum.any?(conclusions, &(&1 == "cancelled")) -> "cancelled"
    Enum.all?(conclusions, &(&1 == "skipped")) -> "skipped"
    true -> "success"
  end
end
```

core/lib/sykli/github/dispatcher.ex
```diff
@@ -125,7 +125,7 @@ defmodule Sykli.GitHub.Dispatcher do
   end
 
   defp acquire_source(event, token, opts) do
-    case source_client(opts).acquire(event, token, opts) do
+    case Sykli.GitHub.Source.acquire(event, token, opts) do
```
yairfalse, line 128: **Naming convention inconsistency in the opts vocabulary.**

The dispatcher's keyword opts mix two patterns now:
- Namespaced: `:app_client`, `:checks_client` (still used by `app_client(opts)` / `checks_client(opts)` helpers)
- Generic: `:impl` (read by the Source facade)

Flat opts make `:impl` ambiguous — it scopes to Source today, but if any other client ever reads `:impl` (e.g., a future `Checks` facade following the same pattern), tests passing `impl: SomeFake` would silently affect both. Right now Source is the only reader so it's safe in practice.

Two ways out, both fine:
- **Keep namespaced everywhere:** rename Source's `:impl` to `:source_impl` (consistent with `:app_client`, `:checks_client`).
- **Keep generic:** standardize all the GitHub clients on facade-level `:impl` resolution and let the dispatcher just call `App.installation_token(opts)`, `Checks.create_run(opts)` etc. without the dispatcher-level helpers.

Not blocking — flagging the convention drift so it gets resolved before more facades land.

core/test/sykli/github/dispatcher_test.exs
```diff
@@ -32,7 +32,7 @@ defmodule Sykli.GitHub.DispatcherTest do
              Dispatcher.dispatch(event,
```
yairfalse, line 33: **Test coverage gap for the new `conclusion/1` branches.**

The diff renames `source_client → impl` here, but doesn't add any test exercising the changed `conclusion/1` mapping (lines 333-342 in `dispatcher.ex`). Specifically untested:

- All-`:skipped` results → suite conclusion `"skipped"`
- Mix of `:passed` + `:blocked` → suite conclusion `"cancelled"` (regression from old `"failure"`)
- Empty results → currently `"skipped"` per the bug above; whatever you decide, lock it in with a test

These are easy to add — construct a fake results list, call `dispatch` end-to-end, assert the `github_check_suite_concluded` occurrence's `conclusion` field. Or unit-test `conclusion/1` directly if you make it `defp` → `@doc false def` for testability.

Without these, a future refactor could silently revert the new behavior.

yairfalse (commented):
## Review summary

Small, well-scoped cleanup PR. Four real changes:

1. **Source facade unification** — collapses two indirection layers (dispatcher's `source_client(opts)` helper + Source's internal `:impl` resolution) into one. Cleaner.
2. **Path segment `:` removal** — `safe_segment` now strips `:` from temp dir names. Real cross-platform fix (`:` is a path separator on Windows; HFS legacy on macOS). Source of `:` is the `"github:#{delivery_id}"` run_id format.
3. **`conclusion/1` derived from `CheckRunFormatter.conclusion/1`** — DRYs the per-task → suite mapping. **But this is a behavior change, not just a refactor** — see inline.
4. **Test improvements** — receiver retry test replaces `eventually` polling with explicit `:receiver_dispatch_completed` signal (genuinely deterministic now); path-containment test gets `unique_integer` + `on_exit`.

### Real concerns

1. **`conclusion/1` behavior change isn't in CHANGELOG** — see inline on `dispatcher.ex:333`.
2. **Empty results map to "skipped" via `Enum.all?([], _)` vacuous truth** — see inline on `dispatcher.ex:339`.
3. **`:impl` vs namespaced opts keys** — naming convention inconsistency. See inline on `dispatcher.ex:128`.
4. **No tests for the new conclusion branches.**

### What's good

- The receiver test rewrite is exactly the right shape — no more `Process.sleep` polling, explicit signal from the test double after dispatch completes
- `unique_integer([:positive])` + `on_exit` for the path-containment test prevents inter-test collision and leftover artifacts on failure
- The added `refute String.contains?(path, ":")` assertions are a clean way to lock in the new safe_segment behavior
- The Source facade now has one canonical injection point (`:impl`) instead of two (the now-removed `source_client(opts)` + the existing `:impl`)

### Recommendation

Approve with concerns 1 and 4 addressed (CHANGELOG entry + tests for the new conclusion branches). Concern 2 is a small bug worth fixing in the same PR. Concern 3 is a convention question that can land separately.

core/lib/sykli/github/dispatcher.ex
yairfalse: Fixed in aab078d. Added an explicit suite_conclusion([]) -> "success" clause so the all-skipped branch cannot catch empty results via vacuous truth.

core/lib/sykli/github/dispatcher.ex
yairfalse: Fixed in aab078d. Added an Unreleased CHANGELOG Changed entry documenting that GitHub check-suite occurrence conclusions now follow the per-task Checks API conclusion mapping.

core/lib/sykli/github/dispatcher.ex
yairfalse: Addressed in aab078d. Source now has :source_impl as the canonical namespaced injection key while preserving :impl as a compatibility fallback for existing direct Source callers.

core/test/sykli/github/dispatcher_test.exs
yairfalse: Fixed in aab078d. Added direct coverage for suite_conclusion/1: empty results -> success, all skipped -> skipped, passed+blocked -> cancelled, and failed+blocked -> failure.

PR #164, opened 2026-05-04
release: prep v0.6.1 with cross-SDK conformance fixes

yairfalse:
## Summary

Bumps all SDK manifests to **0.6.1** and closes **6 of the 11** `expected_failure` cases from the Phase 2.5 burndown by fixing two real cross-SDK conformance bugs that had been masked by the runner's lack of environmental awareness.

**Diff:** 12 files, +68 / −33.

### 🐛 Go SDK — typed-nil leak through `any` field
`sdk/go/sykli.go`. PR #140 changed `jsonTask.Outputs` from `map[string]string` to `any` to accommodate review nodes' `[]string` outputs. But `omitempty` on an interface only strips when the **interface itself** is nil — not when it wraps a typed-nil.

The emitter was building `var outputs map[string]string` (typed nil) and assigning it into `Outputs any`. The interface ended up non-nil but wrapped a nil map — which marshals to `"outputs": null` instead of being omitted. TypeScript and Rust SDKs omit the field entirely; Go was emitting null. Cross-SDK conformance failed for **all 6 SDK-001..006 cases**.

**Fix:** assign `Outputs` only when `len(t.outputs) > 0`. The interface stays at untyped-nil for empty-outputs tasks, omitempty fires correctly. Same fix for review nodes.

### 🐛 Elixir SDK — Logger output corrupting stdout JSON
The SDK has `Logger.debug` calls in `dsl.ex` and `emitter.ex` that fire during `mix run fixture.exs --emit`. Default Logger config writes to stdout, contaminating the JSON output. Fixture emit was producing:

```
23:18:39.262 [debug] registered task
23:18:39.263 [debug] setting timeout
{"version":"1","tasks":[...]}
```

Three lines of garbage before the JSON. Python's json parser failed and the conformance runner crashed.

**Fix:** SDK now ships `config/config.exs` setting Logger level to `:warning` and routing console output to `:standard_error`. Library consumers can override in their own config.

## Conformance runner — env-aware Python skip

`tests/conformance/run.sh` now detects whether the local Python satisfies the SDK's >=3.12 requirement. If not, Python cases are **skipped** instead of **failed**. Python 3.9 (default on macOS) can't even import the SDK because of 3.10+ syntax in the source — that's environmental, not a bug, and shouldn't break the runner.

CI with Python 3.12+ runs Python normally; local dev with older Python sees clean skips.

## Cleanup

- `sdk/typescript/package.json`: `repository.url` now has the `git+` prefix npm wants (fixes the warning on every `npm publish --dry-run`)
- 6 of 11 `expected_failure` flags removed from `dataset.json`: SDK-001..SDK-006 pass naturally. Remaining 5 (CACHE-006, ABN-015, DET-003, DET-006, GH-004) are unrelated bug categories out of scope.

## Test plan

- [x] Conformance: 104 passed, 0 failed, 26 skipped (Python only, env)
- [x] Blackbox: 150 cases, 0 failures
- [x] Core tests: 1,457 / 0 failures
- [x] Go SDK tests passing
- [x] Rust SDK tests: 13/13
- [x] TypeScript SDK tests: 70/70 vitest
- [x] Elixir SDK tests: 38/38
- [x] `mix credo`: clean
- [ ] CI confirms Python SDK + conformance also green with Python 3.12+

## After merge

Run `make release VERSION=0.6.1` from main to bump+commit+tag, then `make publish VERSION=0.6.1` to push to all 5 registries (Go via tag, Rust via cargo, TypeScript via npm, Python via PyPI, Elixir via hex). Tokens required: `CARGO_REGISTRY_TOKEN`, `NPM_TOKEN`, `PYPI_API_TOKEN`, `HEX_API_KEY`.


sdk/go/sykli.go
```diff
@@ -2233,22 +2227,29 @@ func (p *Pipeline) EmitTo(w io.Writer) error {
 				}
 			}(),
 			Verify: t.verify,
-		})
+		}
+		if len(t.outputs) > 0 {
```
yairfalse, line 2231: **Fix is correct — the typed-nil-into-`any` interface gotcha is a textbook Go bug.**

For the curious reader: this works because Go interface values are a `(type, value)` pair. A `var outputs map[string]string` assigned into an `any` field gives `(*map[string]string, nil)` — the interface is non-nil (has type info), so `omitempty` doesn't fire. By only assigning when `len > 0`, the field stays at the zero value `(nil, nil)` and `omitempty` correctly strips it.

**Worth a quick audit of other `any` fields in `jsonTask`.** I can see `Outputs` here; if there are other interface-typed fields (or pointer-to-struct fields whose helper functions return typed-nil under "empty" conditions), they'd have the same bug. Conformance currently passes after this fix, so any latent ones aren't being exercised by the test suite — but that just means the next field added with this pattern will silently regress.

A defensive convention: for any field declared as `any` or `*Foo` in the JSON struct, never assign typed-nil. Either set it conditionally (this PR's fix) or use a typed field with explicit `omitempty`-friendly types (`[]string` instead of `any`).

sdk/elixir/config/config.exs
```diff
@@ -0,0 +1,8 @@
+import Config
+
+# The SDK is invoked via `mix run` to emit JSON to stdout. Logger output
+# would corrupt that JSON, so Logger is silenced by default and routed to
+# stderr if it ever does fire (e.g., a warning during validation). Users
+# embedding the SDK in another app can override these in their own config.
+config :logger, level: :warning
```
yairfalse, line 7: **Library-level Logger config is surprising for SDK consumers.**

This silences Logger globally and reroutes console output to stderr. For the SDK's primary use case — `mix run sykli.exs --emit` from a sykli pipeline — this is exactly right and fixes a real bug.

For consumers who embed the SDK as a dep in their own Elixir app:
- `Mix.install({:sykli_sdk, ...})` since Elixir 1.12 doesn't load deps' `config/config.exs`, so they're safe.
- `mix deps.compile` from a project that depends on `sykli_sdk` also doesn't load this config at the consumer's runtime.
- BUT: any tooling that does load deps' configs (some umbrella setups, some test runners) would silence the consumer's Logger.

The in-file comment says "Users embedding the SDK in another app can override these in their own config" — that's correct mitigation, but the surprise is silent: nothing logs a warning about the override.

**Two stronger alternatives worth considering** (out of scope for this PR but worth a follow-up):

1. **Drop the `Logger.debug` calls in `dsl.ex` and `emitter.ex` entirely.** They're emit-path code — they shouldn't be logging at all. Use `IO.puts(:stderr, ...)` if you really want emitter trace output.
2. **Use a custom logger backend** scoped to the SDK's namespace, instead of changing global Logger config.

For this PR, the current fix is acceptable. Worth at least mentioning in the SDK README / CHANGELOG that consumers may want to be aware.

tests/conformance/run.sh
```diff
@@ -38,6 +38,26 @@ while [[ $# -gt 0 ]]; do
...
+# Detect whether the local Python (with sdk venv preference) satisfies the
+# Python SDK's >=3.12 requirement. If not, the Python conformance cases get
+# skipped instead of failing — env-only gaps shouldn't break the runner.
+SKIP_PYTHON=0
+PYTHON_VERSION="$(
+  cd "$ROOT/sdk/python" 2>/dev/null || exit
+  source .venv/bin/activate 2>/dev/null || true
+  python3 -c 'import sys; print(f"{sys.version_info[0]}.{sys.version_info[1]}")' 2>/dev/null
+)"
+if [[ -z "$PYTHON_VERSION" ]]; then
```
yairfalse, line 50: **Env-aware skip is the right call, but it creates a local-dev coverage gap worth naming.**

The semantics now are: if local Python is <3.12, Python SDK conformance is skipped silently. CI (with Python 3.12+) still catches Python SDK bugs. But a developer running conformance locally on macOS-default Python 3.9 will see "104 passed, 0 failed, 26 skipped" and trust that the runner is happy — even if a real Python SDK bug exists.

**Two small mitigations:**

1. **Print the Python version at the top of the run output**, not just inline per-skip:
   ```bash
   if [[ "$SKIP_PYTHON" -eq 1 ]]; then
     echo "⚠ Python SDK cases will be skipped (local Python: $PYTHON_VERSION; SDK requires >=3.12)"
     echo "  CI runs Python; local devs need Python 3.12+ to exercise SDK conformance locally."
   fi
   ```
   So the developer sees the gap at the start, not just buried in N skip lines.

2. **Exit with a distinct status (e.g., 2)** when Python was skipped due to env, so CI can fail loudly if Python ever ends up unintentionally skipped on the CI runner. Currently the script presumably exits 0 if all non-skipped tests pass — meaning if CI's Python install regresses, the runner silently goes from "104 passed, 0 skipped" to "78 passed, 26 skipped" and CI is still green.

Mitigation 2 is the more important one — it prevents "oh, all Python tests started skipping in CI six weeks ago and we didn't notice."

core/lib/sykli/github/workspace_janitor.ex
```diff
@@ -0,0 +1,51 @@
+defmodule Sykli.GitHub.WorkspaceJanitor do
```
yairfalse, line 1: **This entire file is from open PR #162** (`fix/github-dispatcher-crash-cleanup`). All concerns from my review of #162 still apply and are not addressed in this PR:

1. **Race window upstream of `dispatch_from_source`** — `acquire_source` runs before the janitor exists; a kill in that window leaks the workspace.
2. **Silent 5s timeout in `cleanup/1`** — returns `:ok` on timeout without logging; caller proceeds thinking cleanup happened.
3. **`Task.start` error branch leaks the source** — if it ever fires, no fallback cleanup.
4. **`rescue` doesn't catch Erlang exits** — only Elixir exceptions.
5. **Test only covers crash path; normal-exit cleanup is implicit.**

Full review: https://github.com/false-systems/sykli/pull/162#pullrequestreview-4216980571

If this PR ships v0.6.1 with the janitor included, those concerns ship with it. Either land #162 (with whatever fixes it gets in review) first and rebase this PR, or address those findings here before tagging.

yairfalse (commented):
## Review summary

A release-prep PR that does meaningful work — two real cross-SDK bugs found and fixed (Go typed-nil leak, Elixir Logger contamination), 6 `expected_failure` flags removed naturally, conformance runner becomes env-aware. Good outcomes.

But **two coordination/release-hygiene concerns that should land before merge:**

### 1. This PR bundles all of open PR #162 (workspace janitor) without disclosure

The diff includes the entire `core/lib/sykli/github/workspace_janitor.ex` file, the dispatcher.ex changes, and the dispatcher_test.exs new test from PR #162 — which is currently `OPEN`, not merged. The PR description doesn't mention #162.

Consequences:
- Reviewers reading the description see "version bumps + 2 SDK fixes + dataset cleanup + conformance runner change" and are confused by ~190 lines of unrelated diff.
- All concerns I raised on #162's review (race window upstream of `dispatch_from_source`, silent 5s timeout in `cleanup/1`, `Task.start` error branch leaking source, etc.) carry forward into this release. None are addressed in this PR.
- If #162 merges first, this PR needs a rebase. If this PR merges first, #162 becomes obsolete and its review thread dies orphaned.

**Suggested:** either merge #162 first and rebase this PR (cleanest), or update the description to explicitly note the bundling and link to #162's review thread so its findings travel with the release.

### 2. Missing CHANGELOG entry

Per CLAUDE.md and the Keep-a-Changelog convention, every release needs a `## [0.6.1]` section. This is the release prep PR, so it's the natural place to land it. Should document at minimum:
- Go SDK: typed-nil leak fix for `Outputs` (cross-SDK conformance regression from #140)
- Elixir SDK: Logger silenced + routed to stderr to prevent JSON corruption
- Workspace janitor (assuming #162 work stays bundled)
- Conformance runner: Python cases skipped when local Python <3.12
- 6 `expected_failure` flags removed naturally

### Inline findings

- **`sdk/go/sykli.go:2231`** — fix is correct; worth checking other `any`-typed fields for the same pattern
- **`sdk/elixir/config/config.exs:7`** — library config affects consumers; document the override path more loudly
- **`tests/conformance/run.sh:50`** — env-aware skip is good but creates a local-dev coverage gap worth naming
- **`core/lib/sykli/github/workspace_janitor.ex:1`** — concerns from PR #162 review carry over

### What's good

- Both SDK bugs are real (the Go one masked 6 conformance cases, the Elixir one broke the conformance runner entirely on Python 3.9)
- Removing `expected_failure` flags as bugs are fixed naturally — exactly the pattern CLAUDE.md prescribes
- TypeScript `git+` prefix cleanup eliminates a recurring `npm publish --dry-run` warning
- Test plan in PR description is thorough (10 explicit checks)
- Version-bump scope is clean (all 5 SDKs + Cargo.lock + lockfile + VERSION file all match)

### Recommendation

Do not merge as-is. Either:
1. **Land PR #162 first**, rebase this PR (cleanest — the janitor changes get reviewed and merged through their own thread, this PR shrinks meaningfully).
2. **Update the description** to explicitly disclose the #162 bundle and link to its review, AND add the CHANGELOG entry, AND address the inline findings.

Option 1 is preferable. The release-prep PR should be just: version bumps + the two SDK fixes + dataset cleanup + conformance runner change + CHANGELOG. ~80-100 lines of diff. That's much easier to bisect if anything goes sideways post-release.

PR #167, opened 2026-05-05
docs(schema): add canonical pipeline schema (Phase 2B)

yairfalse:
## Summary

Pins the current Sykli pipeline JSON contract — derived from the engine parser, validation logic, SDK emitters, and the 21 conformance fixtures — in two artifacts. Pure documentation. No source, no schema-emit, no SDK API, no engine behavior changes.

| File | Lines | What it is |
|---|---|---|
| `schemas/sykli-pipeline.schema.json` | 356 | JSON Schema (draft 2020-12), strict (`additionalProperties: false`), describes the canonical SDK-emitted shape |
| `docs/sdk-schema.md` | 298 | Human-readable spec with stability labels, validation boundaries, and known contract gaps |
| `.gitignore` | +1 | `!docs/sdk-schema.md` exception (matches existing `!docs/runtimes.md` pattern) |

## Schema boundary

- **Strict for SDKs:** the canonical schema rejects unknown fields. SDK output should always validate clean.
- **Honest about the engine:** the docs explicitly note three places where the engine is more permissive than the schema — unknown keys silently ignored, `condition` aliases `when`, list-form `outputs` is accepted and normalized. None of these are removed; they're documented.

## Stability labels

All 30 task fields enumerated in Phase 2A are included with explicit stability:

- **stable**: name, command, container, workdir, env, inputs, outputs, depends_on, task_inputs, when, secrets, secret_refs, matrix, services, mounts, retry, timeout, k8s, requires, provides, needs, semantic, ai_hooks, gate
- **deprecated alias**: `condition` (engine accepts; SDKs emit `when`)
- **advisory / unstable**: `target` — emitted by all 5 SDKs, not read by `graph.ex`
- **advisory / reserved**: `version` — emitted by SDKs, not branched on by engine
- **reserved**: `verify`, `oidc` — engine-readable; no SDK emits
- **engine-internal / readOnly**: `history_hint`
- **experimental / Go-only**: `kind: \"review\"` plus `primitive`, `agent`, `context`, `deterministic`

Phase 3 fields (`task_type`, structured inputs/outputs, `success_criteria`, `side_effects`, `expected`) are explicitly listed in `docs/sdk-schema.md` §\"Out of scope\".

## Validation results

All 21 conformance fixtures (`tests/conformance/cases/*.json`) validate cleanly against the new schema:

\`\`\`
PASS  01-basic.json     PASS  08-k8s.json              PASS  15-diamond-deps.json
PASS  02-capabilities   PASS  09-target-requires       PASS  16-long-chain
PASS  03-semantic       PASS  10-task-inputs           PASS  17-outputs
PASS  04-gate           PASS  11-full-options          PASS  18-edge-provides-empty
PASS  05-matrix         PASS  12-minimal               PASS  19-edge-zero-retry
PASS  06-services       PASS  13-kitchen-sink          PASS  20-edge-duplicate-after
PASS  07-secret-refs    PASS  14-gate-file             PASS  21-edge-outputs-v1

Total: 21 pass, 0 fail
\`\`\`

Validated using \`jsonschema\` 4.25.1 from the system Python — **not added as a project dependency**.

## Why no validation script

A \`scripts/validate-conformance-schema.py\` helper was considered and **omitted**: \`jsonschema\` is not in any project dep file (`sdk/python/pyproject.toml` only ships pytest+hypothesis). Adding it requires deciding *where* the dep lives (Python SDK \`[dev]\` extras? new top-level \`requirements-dev.txt\`? Make target with install instructions?), and that decision belongs to Phase 2C. Per the Phase 2B prompt's no-dependency-yak-shaving rule, omission is the right move.

## Stack

Phase 1: PR #165 — accurate docs and conformance Python detection.
Phase 2A: read-only contract discovery (no PR; report inline in conversation).
**Phase 2B: this PR — canonical schema + spec.**
Phase 2C: future — schema decisions on \`version\`/\`target\`/review-node parity, conformance gaps, validation tooling.
Phase 3: future — agent-native semantics (out of scope for the schema).

## Test plan

- [x] \`python3 -c \"import json; json.load(open('schemas/sykli-pipeline.schema.json'))\"\` — schema is well-formed
- [x] All 21 conformance fixtures validate against the schema (jsonschema 4.25.1)
- [x] \`bash tests/conformance/run.sh --sdk python\` — pre-existing skip behavior unchanged on main
- [x] \`git diff origin/main -- core/ sdk/\` is empty — no source changes
- [x] \`git status\` clean

### DIFF
--- a/schemas/sykli-pipeline.schema.json
+++ b/schemas/sykli-pipeline.schema.json
@@ -54,6 +54,25 @@
         {
           "description": "`when` and `condition` are aliased by the engine (`when` wins). SDKs emit `when`. The schema rejects payloads that set both.",
           "not": { "required": ["when", "condition"] }
+        },
+        {
+          "description": "Review-only fields (primitive, agent, context, deterministic) are meaningful only when kind == \"review\". The schema rejects them on regular tasks (where kind is absent or not \"review\").",
+          "if": {
+            "not": {
+              "properties": { "kind": { "const": "review" } },
+              "required": ["kind"]
+            }
+          },
+          "then": {
+            "not": {
+              "anyOf": [
+                { "required": ["primitive"] },
+                { "required": ["agent"] },
+                { "required": ["context"] },
+                { "required": ["deterministic"] }
+              ]
+            }
+          }
         }
       ],
       "properties": {
--- a/schemas/sykli-pipeline.schema.json
+++ b/schemas/sykli-pipeline.schema.json
@@ -1,0 +1,356 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://sykli.dev/schemas/sykli-pipeline.schema.json",
+  "title": "Sykli Pipeline",
+  "description": "Canonical schema for Sykli pipeline JSON as emitted by the five SDKs (Go, Rust, TypeScript, Elixir, Python) and consumed by the engine. This schema describes the SDK-emitted shape and is intentionally strict (additionalProperties: false). The engine is currently more permissive — it ignores unknown fields and accepts a `condition` alias for `when` and a legacy list form for `outputs`. SDKs should not rely on those permissive paths. See docs/sdk-schema.md for the human-readable specification, contract gaps, and stability labels.",
+  "type": "object",
+  "required": ["version", "tasks"],
+  "additionalProperties": false,
+  "properties": {
+    "version": {
+      "description": "Pipeline schema version. Currently advisory — the engine does not branch on this value. SDKs auto-detect: '2' if any container, mount, dir, or cache resource is used; '1' otherwise.",
+      "enum": ["1", "2"]
+    },
+    "tasks": {
+      "description": "List of task objects. May be empty.",
+      "type": "array",
+      "items": { "$ref": "#/$defs/task" }
+    },
+    "resources": {
+      "description": "Map of resource id to resource definition. Only emitted by SDKs when v2 features are present (containers, mounts, dirs, caches).",
+      "type": "object",
+      "additionalProperties": { "$ref": "#/$defs/resource" }
+    }
+  },
+  "$defs": {
+    "resource": {
+      "description": "A directory or cache resource referenced by id from `task.mounts[].resource`.",
+      "type": "object",
+      "required": ["type"],
+      "additionalProperties": false,
+      "properties": {
+        "type": { "enum": ["directory", "cache"] },
+        "path": {
+          "description": "Host path. Required for type=directory.",
+          "type": "string"
+        },
+        "globs": {
+          "description": "Optional glob patterns scoping a directory resource.",
+          "type": "array",
+          "items": { "type": "string" }
+        },
+        "name": {
+          "description": "Cache name. Required for type=cache.",
+          "type": "string"
+        }
+      }
+    },
+    "task": {
+      "description": "A pipeline task (or review node when `kind == \"review\"`). The engine validates that non-gate, non-review tasks have a non-empty `command`; that rule is encoded in description rather than schema-level conditionals because gates and reviews compose with several other fields. SDKs validate cycles, duplicate names, and unknown dependencies; the schema does not.",
+      "type": "object",
+      "required": ["name"],
+      "additionalProperties": false,
+      "allOf": [
+        {
+          "description": "`when` and `condition` are aliased by the engine (`when` wins). SDKs emit `when`. The schema rejects payloads that set both.",
+          "not": { "required": ["when", "condition"] }
+        }
+      ],
+      "properties": {
+        "name": {
+          "description": "Unique task name within the pipeline. Validated non-empty by the engine (validate.ex:115-128) and SDKs.",
+          "type": "string",
+          "minLength": 1
+        },
+        "kind": {
+          "description": "Task kind. Default is task. The review kind is currently emitted only by the Go SDK; engine support is complete but cross-SDK coverage is incomplete. Experimental.",
+          "enum": ["task", "review"]
+        },
+        "command": {
+          "description": "Shell command to execute. Required for regular (non-gate, non-review) tasks; the engine raises `missing_command` otherwise (validate.ex:230-253). Gates and review tasks have no command.",
+          "type": "string"
+        },
+        "container": {
+          "description": "Container image (e.g., `golang:1.21`). Triggers v2 emission.",
+          "type": "string"
+        },
+        "workdir": {
+          "description": "Working directory inside the container.",
+          "type": "string"
+        },
+        "env": {
+          "description": "Environment variables. Keys must be non-empty (validated SDK-side; engine accepts any).",
+          "type": "object",
+          "additionalProperties": { "type": "string" }
+        },
+        "inputs": {
+          "description": "File patterns used for input-based caching. Glob syntax.",
+          "type": "array",
+          "items": { "type": "string" }
+        },
+        "outputs": {
+          "description": "Named output artifacts (output name → path). Engine also accepts a legacy list form which it normalizes into `output_0`, `output_1`, etc., but SDKs emit the map form canonically.",
+          "type": "object",
+          "additionalProperties": { "type": "string" }
+        },
+        "depends_on": {
+          "description": "Names of tasks this task depends on. Engine deduplicates (graph.ex:381) and validates references (validate.ex:159-185).",
+          "type": "array",
+          "items": { "type": "string" }
+        },
+        "task_inputs": {
+          "description": "Cross-task artifact dependencies. Engine validates that `from_task` exists, `output` is declared on it, and the source task is a transitive dependency (graph.ex:766-801).",
+          "type": "array",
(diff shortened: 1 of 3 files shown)


schemas/sykli-pipeline.schema.json
```diff
@@ -0,0 +1,375 @@
...
+          "description": "Environment variables. Keys must be non-empty (validated SDK-side; engine accepts any).",
+          "type": "object",
+          "additionalProperties": { "type": "string" }
+        },
+        "inputs": {
+          "description": "File patterns used for input-based caching. Glob syntax.",
+          "type": "array",
+          "items": { "type": "string" }
+        },
+        "outputs": {
```
yairfalse, line 110: **`outputs` doesn't admit the review-node array form — Go SDK emits arrays for `kind: "review"`.**

In `sdk/go/sykli.go`:
- Line 533: regular tasks: `outputs map[string]string`
- Line 567: review nodes: `outputs []string`
- Line 2005: serialized via `Outputs any` (with `omitempty`)

So the wire shape for a Go review node is:
```json
{ "name": "review/api-breakage", "kind": "review", "outputs": ["api-review.json"] }
```

With the current schema (`outputs: { type: object }`), this would be rejected. The 21 conformance fixtures don't exercise review nodes, so the 21/21 validation result doesn't catch this — but as soon as a review fixture is added, it'll fail.

The spec docs.md acknowledges "Cross-SDK parity for review nodes is a future work item — not part of this schema's scope" and "all five SDKs emit the map form" — but the second claim is incorrect (Go review nodes emit arrays), and the first claim doesn't excuse the schema rejecting the actual wire shape Go SDK currently produces.

**Three options, in increasing order of correctness:**

1. **Document the gap** — add to "Known contract gaps" section: "Review nodes emit `outputs` as `[]string`; non-review tasks emit as `{name → path}` map. Schema currently encodes only the map form; review fixtures will fail validation until cross-SDK parity is reached." Quick + honest, but the schema still misrepresents reality.

2. **Conditional shape** — extend the `if/then` for review-only fields:
```json
"if": { "properties": { "kind": { "const": "review" } }, "required": ["kind"] },
"then": {
  "properties": {
    "outputs": {
      "type": "array",
      "items": { "type": "string" }
    }
  }
}
```
More accurate; encodes that review and non-review have different `outputs` shapes.

3. **Fix Go SDK to emit map form for reviews too** — likely the right long-term answer (review outputs become `{0: "api-review.json"}` or `{result: "api-review.json"}`), but it's an SDK breaking change and out of scope for a docs PR.

Option 1 in this PR + an explicit follow-up tracking 2 or 3 is probably the right scope here.

schemas/sykli-pipeline.schema.json
```diff
@@ -0,0 +1,375 @@
...
+            "client_id": { "type": "string" },
+            "duration": {
+              "description": "Token duration in seconds. Engine default 3600.",
+              "type": "integer",
+              "minimum": 1
+            },
+            "audience": { "type": "string" }
+          }
+        },
+        "history_hint": {
```
yairfalse, line 350: **`history_hint` in `properties` means SDKs CAN emit it — contradicts the spec's "MUST NOT emit".**

Under `additionalProperties: false`, fields listed in `properties` are *the* allowed fields. So this entry is what *permits* `history_hint` to appear in SDK output, not what forbids it. The `readOnly: true` annotation is informational for OpenAPI-style tooling but isn't enforced by JSON Schema validators at validate-time.

The spec docs.md is unambiguous: "SDKs MUST NOT emit this field." If that's the actual contract, the schema should enforce it.

**Two ways to actually forbid SDK emission:**

1. **Remove from `properties` entirely.** If SDKs aren't supposed to emit it, the schema (which describes "the canonical SDK-emitted shape") shouldn't list it. The engine reads it but doesn't need a schema entry for that.

2. **Active rejection via `not`** in the task `allOf`:
```json
{
  "description": "history_hint is engine-internal; SDKs must not emit it.",
  "not": { "required": ["history_hint"] }
}
```

Option 1 is cleaner. The current state (listed + readOnly) is the worst of both worlds: it tells SDK authors "you may emit this" but tells human readers "you must not."

The "parser symmetry" justification in the description is for the engine, but the schema isn't the engine's input — it's the SDK's output contract.

schemas/sykli-pipeline.schema.json
```diff
@@ -0,0 +1,375 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://sykli.dev/schemas/sykli-pipeline.schema.json",
```
yairfalse, line 3: **Schema `$id` has no version segment — every future change is implicitly breaking for URL-based consumers.**

The URL `https://sykli.dev/schemas/sykli-pipeline.schema.json` will always point to whatever's currently published. If the schema later adds a required field, removes a stable enum value, or tightens a shape, every tool that fetched by `$id` now silently gets a different contract.

**Suggested:** version the path:
- `https://sykli.dev/schemas/v1/sykli-pipeline.schema.json` (path-versioned)
- or `https://sykli.dev/schemas/sykli-pipeline.v1.schema.json` (filename-versioned)

Either way, future v2 changes can ship at `/v2/...` while v1 consumers continue to work. This matches how JSON Schema, OpenAPI, and most public schemas evolve.

**Companion concern:** the `sykli.dev` domain doesn't appear to actually serve this content yet (couldn't find a `schemas/` route from the README links). Until it does, consumers fetching by `$id` will 404. Either:
- Stand up the URL (and version it as above), or
- Use a `urn:uuid:...` `$id` so it's clearly identifier-only, not fetchable, or  
- Use a stable GitHub raw URL: `https://raw.githubusercontent.com/false-systems/sykli/main/schemas/sykli-pipeline.schema.json`

All three are defensible; the current state (URL-shaped but non-functional and unversioned) is the only one that's not.

yairfalse (commented):
## Review summary

Exactly the right shape of work, and unusually well-grounded for a schema PR. The descriptive-not-prescriptive framing, the source-line references throughout (`graph.ex:382`, `validate.ex:115-128`, etc.), the explicit "engine is more permissive than the schema in three named ways" section, and the 8-item "Known contract gaps" all make this maintainable rather than instantly stale.

21/21 conformance fixtures validate cleanly. Pure docs (no code, no CI). The "why no validation script" justification is the right call for now — pinning the dep belongs to Phase 2C.

### Real concerns

1. **`outputs` schema doesn't admit the review-node array form** — Go SDK emits `[]string` for review-node outputs, schema only accepts object. Genuine schema-vs-reality mismatch. See inline on line 110.
2. **`history_hint` in `properties` actually permits SDKs to emit it** — under `additionalProperties: false`, listing a field in `properties` is what allows it. Contradicts the spec's "SDKs MUST NOT emit." See inline on line 350.
3. **Schema `$id` has no version** — any future schema change is implicitly breaking for any tool that fetches by URL. See inline on line 3.
4. **No CI enforcement of schema-vs-fixture parity** — the 21/21 validation is a one-shot author check. Future SDK or schema drift goes unnoticed until someone runs the validator manually. (Body-level finding; resolution is Phase 2C per the PR description.)

### Smaller findings

- **The `if/then` for review-only fields** (lines 58-77) is logically correct but dense — `if/then/else` would be easier to read.
- **Gate `oneOf` could enforce conditional `env_var` / `file_path` / `webhook_url` requirements per strategy.** Currently all gate fields are independently optional even though the engine requires them per-strategy.
- **`oidc.provider` enum is closed-set** (`aws`/`gcp`/`azure`). If new IdPs are added, every consumer of the schema needs an update. Open-string with a `pattern` may age better.
- **`docs/sdk-audit.md` is referenced in the spec but is gitignored** — readers can't see it. Worth a clarifying note that the audit doc is internal-only and won't appear in checkouts.
- **Worth referencing `docs/sdk-schema.md` from `CLAUDE.md`** so future contributors and Claude sessions discover it.

### What's good

- The contract-boundary section is the kind of honest documentation that prevents future bug reports — it explicitly tells implementers: "if a payload validates schema-clean, the engine accepts it; if a payload fails the schema but the engine accepts it, you're using a permissive path that SDKs shouldn't rely on." That's the right abstraction.
- Source-line references throughout (`graph.ex:368-414`, `validate.ex:115-128`, `history_hint.ex:5`) tie the schema to specific commits' worth of behavior. Future drift is easy to detect.
- Stability labels for every field. The `stable` / `deprecated alias` / `advisory / unstable` / `reserved` / `experimental` taxonomy is exactly what agentic consumers need to reason about pattern-matching durability.
- The "Out of scope" section explicitly names Phase 3 fields (`task_type`, structured `inputs`/`outputs`, `success_criteria`, `side_effects`, `expected`) so reviewers don't ask why they're missing.
- `additionalProperties: false` everywhere — strict by default, exceptions called out. This is the right default for a contract.
- The `not: { required: ["when", "condition"] }` construct correctly forbids both being set even though the engine accepts either.

### Recommendation

Approve with the three inline concerns addressed (`outputs` array form for reviews, `history_hint` removal from properties, `$id` versioning). The smaller findings are nice-to-haves that can land separately or in Phase 2C.

PR #169, opened 2026-05-05
Add canonical SDK schema guardrails

yairfalse:
## Summary
- document the top-level pipeline version as the wire-format/schema version
- remove target from the canonical SDK-emitted contract while keeping builder APIs as deprecated no-ops
- add conformance fixture validation against schemas/sykli-pipeline.schema.json
- fix Python conformance interpreter detection so SYKLI_CONFORMANCE_PYTHON participates and Python >=3.12 is preferred

## Validation
- python3 scripts/validate-conformance-schema.py -> 21 passed, 0 failed
- env GOCACHE=/private/tmp/sykli-go-build SYKLI_CONFORMANCE_PYTHON=python3.14 tests/conformance/run.sh -> 130 passed, 0 failed, 0 skipped
- go test ./... in sdk/go
- cargo test --lib in sdk/rust
- npx vitest run in sdk/typescript
- mix test in sdk/elixir
- PYTHONPATH=src python3.14 -m pytest in sdk/python

## Scope confirmations
- no engine behavior changes
- no replacement field for target
- no Phase 3 fields
- no schema field changes in 2C-3/2C-4


yairfalse (commented):
## Code Review

### Overview
Three coordinated changes:
1. **Documents `version`** as the pipeline wire-format/schema version (not an execution capability selector), with a migration plan to version-aware parsing.
2. **Removes `target`** from the canonical SDK-emitted contract across all five SDKs and the JSON Schema; builder methods are kept as deprecated no-ops, conformance fixtures `09-target-requires` and `13-kitchen-sink` are updated.
3. **Adds `scripts/validate-conformance-schema.py`** and wires it as a pre-flight step in `tests/conformance/run.sh`; also fixes the runner's Python interpreter detection so `SYKLI_CONFORMANCE_PYTHON` is honored and ≥3.12 is preferred.

### Strengths
- **Schema strictness pays off here**: `additionalProperties: false` on the task object means removing `target` from the schema actually rejects it, not just deprecates it. The new validator runs at the start of `run.sh` and catches drift between SDK output and contract before the byte-equality phase.
- **Cross-SDK consistency**: all five SDKs updated in lockstep — Go (`sykli.go:2010, 2155`), Rust (`lib.rs:2255, 2588`), TS (`index.ts:882`), Elixir (`emitter.ex:200`), Python (`__init__.py:743`). Conformance fixtures + golden JSON updated for every SDK. This is exactly the discipline `docs/sdk-schema.md` calls for.
- **Tests updated, not just code**: `test_serialization.py:60` and `test_task.py:247` now assert `"target" not in task` — the negative assertion is the right shape.
- **Versioning doc is good prose**: "minimum pipeline wire-format version required to preserve its meaning" is the correct framing, and the explicit "do not design or reserve the contents of `version: \"3\"` until a concrete wire-format change requires it" guards against speculative API design.
- **Schema validator script is well-formed**: validates schema itself with `Draft202012Validator.check_schema` before validating cases, sorts errors by `absolute_path`, exits 2 on missing dep with install hint.

### Issues & Suggestions

#### Medium

**1. Silent deprecation in Python and Go.** Rust uses `#[deprecated]`, Elixir uses `@deprecated`, TS uses JSDoc `@deprecated` (IDE-visible). But:
- **Python (`sdk/python/src/sykli/__init__.py:579`)**: docstring only. Add a `warnings.warn(..., DeprecationWarning, stacklevel=2)` so users actually find out their `.target(\"k8s\")` call became a no-op. This is the standard Python deprecation pattern and the cost is one line.
- **Go (`sdk/go/sykli.go:1271`)**: only a `// Deprecated:` comment. That *is* the Go convention and `staticcheck` / `go vet` will flag it, so this is acceptable — but worth a project decision.

**2. Go's `Target(\"\")` still panics on a no-op (`sdk/go/sykli.go:1272-1274`).** A function that does nothing should not panic on input validation. Either drop the panic (the function is a no-op, input doesn't matter) or accept the inconsistency consciously. Matches Rust which kept its `assert!(!name.is_empty(), ...)` for the same reason. I'd drop both.

**3. Explicit `SYKLI_CONFORMANCE_PYTHON` can be silently overridden (`tests/conformance/run.sh:62-75`).** If a user sets `SYKLI_CONFORMANCE_PYTHON=python3.10` (which fails the ≥3.12 check), the loop falls through to other candidates and overwrites their explicit choice. Either:
- error out loudly when an explicit override fails the version check, or
- at minimum, `echo` a warning when the explicit value is replaced.

The current behavior is friendlier-to-CI but surprising for local devs debugging interpreter selection.

**4. New hard dependency on `jsonschema` for all conformance runs.** With `set -euo pipefail` and the validator running first, any dev without the `jsonschema` Python package now gets a hard failure on `tests/conformance/run.sh`. The script does print an install hint, but consider:
- documenting it in CLAUDE.md / GETTING_STARTED.md alongside the existing Python ≥3.12 note, and/or
- gating it on the same Python availability check the runner already does (currently the validator runs *before* that check).

#### Low / Nits

**5. Doc duplication.** `docs/sdk-schema.md` now has the `version` story in three places: the `version` field bullet, the new `## Versioning decision` section (50+ lines), and `## Known contract gaps` item 1. Some of this is reasonable (bullet → narrative → known-gap pointer), but the "Versioning decision" section repeats the `\"1\"` vs `\"2\"` definitions verbatim. Consider trimming the section or removing the bullet-level redefinition.

**6. `target` removal note** (`docs/sdk-schema.md:120-130`): "Existing engines may still ignore incoming `target` fields for compatibility because the engine parser is permissive" — this is descriptively accurate but could mislead users into thinking the engine still has affordances for `target`. A one-liner clarifying that the engine never read it in the first place would close that gap.

**7. `scripts/validate-conformance-schema.py:42-49`**: bare `except Exception` for schema/JSON load. Acceptable since failure mode is "give the user a path + message and exit", but `OSError` + `json.JSONDecodeError` would communicate intent better.

**8. CHANGELOG not updated.** This is a user-visible breaking change to SDK output (target field disappears from JSON) — even though the builder method survives, callers who serialize and inspect the JSON will see a difference. Worth a CHANGELOG entry under `Removed` / `Changed`.

### Risk Assessment

- **Backward compat**: Engine is permissive on unknown fields, so old SDK output (with `target`) still parses. New SDK output (without `target`) was already accepted. Zero engine-side risk.
- **Downstream tooling**: anything that grepped emitted JSON for `\"target\":` will break. Low population, but real.
- **Test coverage**: schema validator + conformance byte-equality + per-SDK serialization tests cover this from three angles. Solid.
- **CI**: PR description shows all five SDKs' tests passing plus the new schema validator (21/21).

### Verdict
Approve with the suggestion to add a Python `DeprecationWarning`, decide on the Go panic, and either fail loudly or warn when `SYKLI_CONFORMANCE_PYTHON` gets overridden. The contract cleanup is exactly what `docs/sdk-schema.md` was asking for.

PR #170, opened 2026-05-06
Add experimental review-node SDK support

yairfalse:
## Summary
- add minimal experimental review-node builders/APIs across Go, Rust, TypeScript, Python, and Elixir
- add conformance case 22-review-node plus fixtures for all SDKs
- document review nodes as experimental and keep review outputs out of canonical behavior

## Validation
- python3 scripts/validate-conformance-schema.py — 22 passed, 0 failed
- env GOCACHE=/private/tmp/sykli-go-build SYKLI_CONFORMANCE_PYTHON=python3.14 tests/conformance/run.sh — 135 passed, 0 failed, 0 skipped
- env GOCACHE=/private/tmp/sykli-go-build SYKLI_CONFORMANCE_PYTHON=python3.14 tests/conformance/run.sh 22-review-node — 5 passed
- env GOCACHE=/private/tmp/sykli-go-build go test ./...
- cargo test --lib
- npx vitest run
- mix test
- PYTHONPATH=src python3.14 -m pytest
- git diff --check
- git diff --name-only core schemas — no output

Note: the conformance harness embedded schema-validation step skipped because python3.14 does not have jsonschema installed. Standalone schema validation passed with local python3.

## Confirmations
- no engine behavior changes
- no schema field changes
- no replacement for target
- no review outputs added to canonical behavior
- no Phase 3 fields added

### DIFF
--- a/docs/sdk-schema.md
+++ b/docs/sdk-schema.md
@@ -69,7 +69,7 @@ The full canonical field list, with stability labels:
 | JSON key | Stability | Required? | Notes |
 |---|---|---|---|
 | `name` | stable | yes | unique, non-empty |
-| `kind` | experimental (non-Go SDKs) | no | `"task"` (default) or `"review"`; only Go emits `"review"` |
+| `kind` | experimental | no | `"task"` (default) or `"review"` |
 | `command` | stable | conditional | required unless `gate` is set or `kind == "review"` |
 | `container` | stable | no | triggers v2 |
 | `workdir` | stable | no | |
@@ -108,7 +108,7 @@ Unique within the pipeline, non-empty. The engine rejects empty/whitespace/non-b
 
 ### `kind`
 
-`"task"` (default; SDKs do not emit `kind` in this case) or `"review"`. When `kind == "review"`, `command` is not required and the review-only fields below become meaningful. Today only the Go SDK constructs review nodes (`sdk/go/sykli.go:559-570`, `:2241-2247`); engine support is complete (`graph.ex:420-429`, `parse_review/1`). Cross-SDK parity for review nodes is a future work item — not part of this schema's scope.
+`"task"` (default; SDKs do not emit `kind` in this case) or `"review"`. When `kind == "review"`, `command` is not required and the review-only fields below become meaningful. Review nodes are experimental graph nodes for review/reasoning/check primitives, not shell-command tasks.
 
 ### `command`
 
@@ -242,7 +242,13 @@ When `kind == "review"`, the engine reads four additional fields via `parse_revi
 - `context` — array of context file references.
 - `deterministic` — boolean; engine default `false`.
 
-These are meaningful only when `kind == "review"`. The schema **rejects** them on regular tasks (encoded via `if/then` in the task `allOf`): if `kind` is absent or not `"review"`, none of `primitive`, `agent`, `context`, or `deterministic` may appear. The engine itself is permissive — `parse_review/1` (`graph.ex:420-429`) is simply never entered for non-review tasks, so any review fields would be silently ignored — but the schema treats them as a likely typo. Cross-SDK support is incomplete: only Go currently emits review nodes.
+These are meaningful only when `kind == "review"`. The schema **rejects** them on regular tasks (encoded via `if/then` in the task `allOf`): if `kind` is absent or not `"review"`, none of `primitive`, `agent`, `context`, or `deterministic` may appear. The engine itself is permissive — `parse_review/1` (`graph.ex:420-429`) is simply never entered for non-review tasks, so any review fields would be silently ignored — but the schema treats them as a likely typo.
+
+Review nodes do not have canonical `outputs` behavior yet. SDKs should not emit
+`outputs` for review nodes; review results/structured outputs are intentionally
+left out of the current experimental contract. The schema rejects task execution
+fields on review nodes: `command`, `outputs`, `gate`, `container`, `services`,
+`k8s`, `mounts`, `retry`, and `timeout`.
 
 ## Normalization behavior
 
@@ -352,7 +358,7 @@ introduce a replacement field.
 These are **descriptive, not prescriptive**. The schema documents current behavior; resolving these is Phase 2C / future work.
 
 1. **`version` is advisory only today.** SDKs emit `"1"` or `"2"`; engine ignores. This document defines it as the pipeline wire-format/schema version and describes the migration path to version-aware parsing.
-2. **Review nodes are Go-only.** Engine supports `kind: "review"` and the four review-only fields, but only the Go SDK constructs them. No conformance coverage.
+2. **Review nodes are experimental.** Engine supports `kind: "review"` and the four review-only fields, and SDKs expose minimal review builders. Review outputs are not canonical.
 3. **`verify` and `oidc` are reserved with no SDK emit.** Engine reads them; SDKs have no API. Either implement SDK support or drop from the schema once a decision lands.
 4. **`history_hint` is engine-internal.** SDKs MUST NOT emit. Schema marks `readOnly` for clarity.
 5. **TypeScript K8s interface drift.** `K8sOptions` interface declares 15 fields; only 4 are serialized. The other 11 silently disappear at emit. The schema reflects the wire contract (4 fields); the TypeScript drift is a TS-side issue, not a schema issue.
--- a/schemas/sykli-pipeline.schema.json
+++ b/schemas/sykli-pipeline.schema.json
@@ -73,6 +73,28 @@
               ]
             }
           }
+        },
+        {
+          "description": "Review nodes are graph metadata nodes, not shell-command tasks. The canonical review-node contract rejects task execution fields.",
+          "if": {
+            "properties": { "kind": { "const": "review" } },
+            "required": ["kind"]
+          },
+          "then": {
+            "not": {
+              "anyOf": [
+                { "required": ["command"] },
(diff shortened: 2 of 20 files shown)


yairfalse (commented):
## Overview

This PR brings the Elixir, Rust, TypeScript, and Python SDKs up to the level the Go SDK already had for `kind: "review"` graph nodes — minimal experimental builders for `primitive`, `agent`, `context`, `inputs`, `after`, `deterministic`. It also retires the Go-only special case in `docs/sdk-schema.md`, removes the Go SDK's review-`outputs` emission, and adds conformance case `22-review-node` with fixtures in all five SDKs. The schema, the engine parser, and the canonical contract are intentionally untouched.

## Strengths

- **Cross-SDK coverage actually delivered**, not stubbed. All five SDKs emit byte-equivalent JSON for the new conformance case (135/0/0 in the validation block).
- **Validation parity**: every SDK now raises/panics if `kind: "review"` is set without a `primitive`, mirroring the engine's existing requirement.
- **Schema discipline preserved**: `outputs` removed from review-node emission across SDKs (Go `sykli.go:2242-2244`, Rust `lib.rs:2351`, Elixir's `review_to_json/1`, Python's `Review._to_dict`, TS's `Review._toJSON`). The doc paragraph explicitly states "Review nodes do not have canonical `outputs` behavior yet."
- **Docs cleanup is the real value-add**: removing the "Go-only" caveat from `kind` and the "Review nodes are Go-only" known-gap entry tightens the schema document considerably.
- **Test coverage**: Go has positive + negative + Outputs-noop test; Rust has serialization + no-command + missing-primitive; Python has no-command + missing-primitive; Elixir has serialization + missing-primitive + rejection-of-task-options; TS has serialization + no-command. Real coverage, not just happy-path smoke.

## Issues

### High

**1. Smart-quote regression in `sdk/go/sykli.go:1019`** (unrelated to review nodes — got dragged in by an editor).
```
-//   - tag != '' - run only when a tag is present
+//   - tag != ” - run only when a tag is present
```
The closing `''` got auto-corrected to a Unicode right double quotation mark (`U+201D`). Reads as broken Go-string syntax to anyone reading the docstring. Not a runtime issue (it's a comment), but it's a regression. **Restore `''`** before merge.

**2. Schema does not enforce "no `outputs` on review tasks"** — only docs do. `schemas/sykli-pipeline.schema.json` has an `allOf[1]` rule that rejects `primitive/agent/context/deterministic` on non-review tasks, but no symmetric rule rejecting `outputs` (or `command`, `services`, `mounts`, `k8s`, …) on review tasks. The validator script can't catch a future regression where an SDK or hand-rolled fixture emits `{"kind": "review", "outputs": {…}}`. Docs say SDKs MUST NOT do this; schema permits it.

Either:
- Add a third `allOf` clause: `if kind == "review" then not anyOf [required:["outputs"], required:["command"], required:["services"], required:["mounts"], required:["k8s"], required:["retry"], required:["timeout"]]`, OR
- Acknowledge the gap explicitly in `## Known contract gaps` so future readers know the boundary lives in docs+SDKs+conformance only.

The first is the right fix given that the schema *already* models the inverse rule.

### Medium

**3. Elixir's `reject_review_option!` is asymmetrically applied.** It guards `run`, `container`, `outputs`, `output`, `service`, `retry`, `timeout`, `k8s`, `mount`, `mount_cache`, `mount_cwd`, `mount_cwd_at`, and the `gate_*` family — but not `workdir`, `inputs`, `env`, `secret`, `secret_from`, `matrix`, `requires`, `provides`, `needs`, `covers`, `intent`, `set_criticality`, `on_fail`, `select_mode`, `when_/condition`, `task_input`, `before/after`. A user calling `review "x" do … env("CI", "true") end` silently mutates the review struct in a way that won't serialize but also won't raise. Two paths:

- (a) Cover all task-only setters with `reject_review_option!` for fail-loud parity.
- (b) Drop the rejection logic entirely and rely on `review_to_json/1` simply not emitting fields that don't exist on the review wire shape (matching how Python and TS naturally handle this — separate `Review` class, no shared methods).

Option (b) is cleaner and matches the structural separation other SDKs got for free. Elixir's process-dictionary DSL forced the runtime guard, but a half-implemented runtime guard is worse than none.

**4. `Review.Outputs` in Go is now a no-op-with-input-validation** — the same anti-pattern PR #169 just fixed for `Task.Target()`. `sykli.go:802-812` does nothing with `paths` except panic on empty strings. Either the function is a no-op (in which case input validation is contradictory), or it isn't. Drop the `if path == ""` panic to match the precedent we just set, or accept the inconsistency consciously.

The Rust SDK doesn't have `Review::outputs` at all (cleaner). Python/TS don't either. Go is alone in keeping it for "source compatibility with early experimental users" — fair, but then make it a true no-op.

**5. Inconsistent empty-input handling across SDKs.** Rust's `Review::primitive`/`agent`/`context`/`inputs` all `assert!(!s.is_empty())` — panics. Python raises `ValueError`. TS throws `Error`. Go panics via `log.Panic()`. Elixir raises `ArgumentError`. The choices are language-idiomatic individually, but the Rust panics are notable because PR #169 specifically dropped Rust panics on no-op functions. These setters aren't no-ops, so the analysis differs — but it's worth deciding once whether experimental-API setters should panic or return `Result`. Given everything else in the Rust SDK panics on misuse, current behavior is internally consistent. Flagging for visibility, not blocking.

### Low / Nits

**6. Negative test coverage gap** on the new APIs:
- No SDK tests `.review("")` raises/panics.
- No SDK tests duplicate-name collision (`p.task("x"); p.review("x")`).
- Rust/TS don't test `.primitive("")` raises.

These are exactly the cases the asserts/raises were added to defend, and none of them are exercised.

**7. TS `Review` duck-types `Task`'s internal protocol via `_isReview/_isGate/_getCommand/_getMounts/_getK8s/_setK8s` stubs.** Works, but having `Pipeline.tasks: Array<Task | Review>` and then calling underscore-prefixed methods that exist on both classes is fragile — adding a new `Task._method` would silently miss `Review`. A shared `PipelineNode` interface would tighten this. Refactor scope, not a blocker.

**8. Rust `Review::after` allocates a temp `String` per dedup check** (`task.depends_on.contains(&name.to_string())`). For typical ≤5-element dep lists, irrelevant; but `.iter().any(|d| d == name)` avoids the alloc and reads cleaner.

**9. Python `.deterministic(value)` coerces** with `bool(value)` — accepts truthy non-bool inputs. Other SDKs require an actual boolean. Minor wire-format risk if a caller passes `1` or `"yes"` and expects validation. Not exploitable, just inconsistent.

## Risk Assessment

- **Engine-side risk: zero.** No engine code changed. Engine already supports `kind: "review"` and `parse_review/1`.
- **Schema risk: low but real.** Schema is permissive about `outputs` on reviews; if a future SDK regresses and emits them, schema validation won't catch it. (See issue 2.)
- **Cross-SDK divergence risk: low.** Conformance case 22 anchors byte equality. Asymmetric Elixir validation (issue 3) means the same *misuse* fails differently across SDKs, but emitted JSON for valid use stays consistent.
- **Backward compatibility: clean.** New APIs are additive. Go's removed `Review.Outputs` storage is a breaking change for anyone who relied on it being emitted, but the previous version emitted a field the canonical contract now disclaims; PR #169 already established the same pattern for `target`.
- **Experimental designation honored.** Docs label the feature experimental; nothing leaks into stable schema fields.

## Verdict

**Approve with two requested fixes before merge:**

1. Restore `tag != ''` in `sdk/go/sykli.go:1019` (smart-quote regression).
2. Either add the schema constraint rejecting `outputs/command/services/mounts/k8s` on review tasks (preferred, matches existing inverse rule), or document the gap in `## Known contract gaps`.

- Decide and apply consistent rejection-of-task-options policy in Elixir (cover all setters or drop the runtime guard — issue 3).
- Drop the empty-string panic in Go's `Review.Outputs` (matches precedent from PR #169 — issue 4).
- Add negative tests for `.review("")`, duplicate names, and `.primitive("")` (issue 6).

The contract cleanup of removing the Go-only caveat is the real prize here, and it lands cleanly. Solid PR.

yairfalse:
Addressed the review items and pushed an amended commit.

Changes made:
- Fixed the Go smart quote typo in the tag parsing check.
- Tightened the canonical schema so review nodes reject task execution fields: command, outputs, gate, container, services, k8s, mounts, retry, and timeout.
- Made Go Review.Outputs a true compatibility no-op with no validation side effects.
- Expanded Elixir review guards across the remaining task-only DSL setters.
- Added negative tests for empty review names, duplicate review names, and empty primitives across SDKs where applicable.
- Made Python Review.deterministic require a real bool instead of bool-coercing arbitrary values.
- Fixed the TypeScript typecheck failure around Task | Review ordering/conditions.
- Removed the small Rust allocation in Review.after.

Validation run locally:
- python3 scripts/validate-conformance-schema.py -> 22 passed, 0 failed
- env GOCACHE=/private/tmp/sykli-go-build SYKLI_CONFORMANCE_PYTHON=python3.14 tests/conformance/run.sh -> 135 passed, 0 failed, 0 skipped
- sdk/go: go test ./...
- sdk/rust: cargo test --lib
- sdk/typescript: npx tsc --noEmit; npx vitest run
- sdk/elixir: mix test
- sdk/python: PYTHONPATH=src python3.14 -m pytest
- git diff --check

Note: the conformance harness still reports that embedded schema validation is skipped under python3.14 because that interpreter does not have jsonschema installed. Standalone schema validation passed under local python3, and the harness SDK conformance passed.

PR #172, opened 2026-05-06
feat: add task_type as semantic contract field

yairfalse:
## Summary
Adds `task_type` as the first version 3 semantic contract field.

- `task_type` is executable-task-only semantic metadata
- version `3` is required when `task_type` appears
- enum is the 12 Phase 3B values
- rejected on review nodes
- no `custom`
- no preset inference
- all five SDKs support explicit APIs
- conformance adds `23-task-type` across all SDKs
- core uses a shared `Sykli.TaskType` enum source
- parse-time task_type errors have user-facing formatting

## Validation
- `python3 -c "import json; json.load(open('schemas/sykli-pipeline.schema.json'))"`
- `python3 scripts/validate-conformance-schema.py` -> 23 passed, 0 failed
- schema negative task_type checks -> passed
- `tests/conformance/run.sh` -> 140 passed, 0 failed, 0 skipped
- `mix test` in `core` -> 1476 tests, 0 failures
- `cd sdk/go && GOCACHE=/private/tmp/sykli-go-build go test ./...`
- `cd sdk/rust && cargo test --lib` -> 108 passed
- `cd sdk/typescript && npx vitest run` -> 79 passed
- `cd sdk/typescript && npx tsc --noEmit`
- `cd sdk/python && PYTHONPATH=src python3.14 -m pytest` -> 202 passed
- `cd sdk/elixir && mix test` -> 47 passed

Note: the conformance harness embedded schema-validation step still skips under `python3.14` because that interpreter lacks `jsonschema`; standalone schema validation with local `python3` passed.

### DIFF
--- a/core/lib/sykli/graph.ex
+++ b/core/lib/sykli/graph.ex
@@ -373,15 +373,15 @@ defmodule Sykli.Graph do
   end
 
   def format_error({:task_type_on_review, task_name}) do
-    "Review node '#{task_name}' cannot declare task_type"
+    "Error: Review node '#{task_name}' cannot declare task_type"
   end
 
   def format_error({:task_type_requires_version_3, task_name, version, _task_type}) do
-    "Task '#{task_name}' declares task_type but pipeline version is #{inspect(version)}, not \"3\""
+    "Error: Task '#{task_name}' declares task_type but pipeline version is #{inspect(version)}, not \"3\""
   end
 
   def format_error({:unknown_task_type, task_name, task_type}) do
-    "Task '#{task_name}' declares unknown task_type #{inspect(task_type)}"
+    "Error: Task '#{task_name}' declares unknown task_type #{inspect(task_type)}"
   end
 
   def format_error(reason), do: inspect(reason)
--- a/core/lib/sykli/graph.ex
+++ b/core/lib/sykli/graph.ex
@@ -71,6 +71,7 @@ defmodule Sykli.Graph do
     defstruct [
       :name,
       :kind,
+      :task_type,
       :command,
       :inputs,
       :outputs,
@@ -142,6 +143,10 @@ defmodule Sykli.Graph do
     @spec kind(t()) :: :task | :review
     def kind(%__MODULE__{kind: :review}), do: :review
     def kind(_), do: :task
+
+    @doc "Returns the executable task semantic class."
+    @spec task_type(t()) :: String.t() | nil
+    def task_type(%__MODULE__{task_type: task_type}), do: task_type
 
     @doc "Returns the task command."
     @spec command(t()) :: String.t()
@@ -347,8 +352,10 @@ defmodule Sykli.Graph do
 
   def parse(json) do
     case Jason.decode(json) do
-      {:ok, %{"tasks" => tasks}} ->
-        case map_ok(tasks, &parse_task/1) do
+      {:ok, %{"tasks" => tasks} = data} ->
+        version = Map.get(data, "version", "1")
+
+        case map_ok(tasks, &parse_task(&1, version)) do
           {:ok, parsed_tasks} ->
             parsed = Map.new(parsed_tasks, fn task -> {task.name, task} end)
             {:ok, parsed}
@@ -365,16 +372,32 @@ defmodule Sykli.Graph do
     end
   end
 
-  defp parse_task(map) do
+  def format_error({:task_type_on_review, task_name}) do
+    "Review node '#{task_name}' cannot declare task_type"
+  end
+
+  def format_error({:task_type_requires_version_3, task_name, version, _task_type}) do
+    "Task '#{task_name}' declares task_type but pipeline version is #{inspect(version)}, not \"3\""
+  end
+
+  def format_error({:unknown_task_type, task_name, task_type}) do
+    "Task '#{task_name}' declares unknown task_type #{inspect(task_type)}"
+  end
+
+  def format_error(reason), do: inspect(reason)
+
+  defp parse_task(map, version) do
     task_name = map["name"]
     kind = parse_kind(map["kind"])
 
-    with {:ok, services} <- parse_services(map["services"], task_name),
+    with {:ok, task_type} <- parse_task_type(map["task_type"], kind, version, task_name),
+         {:ok, services} <- parse_services(map["services"], task_name),
          {:ok, mounts} <- parse_mounts(map["mounts"], task_name) do
       {:ok,
        %Task{
          name: task_name,
          kind: kind,
+         task_type: task_type,
          command: map["command"],
          inputs: map["inputs"] || [],
          outputs: normalize_outputs(map["outputs"], kind),
@@ -416,6 +439,24 @@ defmodule Sykli.Graph do
   defp parse_kind("review"), do: :review
   defp parse_kind(:review), do: :review
   defp parse_kind(_), do: :task
+
+  defp parse_task_type(nil, _kind, _version, _task_name), do: {:ok, nil}
+
+  defp parse_task_type(_task_type, :review, _version, task_name) do
+    {:error, {:task_type_on_review, task_name}}
+  end
+
+  defp parse_task_type(task_type, _kind, version, task_name) when version != "3" do
+    {:error, {:task_type_requires_version_3, task_name, version, task_type}}
+  end
+
+  defp parse_task_type(task_type, _kind, "3", task_name) do
+    if Sykli.TaskType.valid?(task_type) do
+      {:ok, task_type}
+    else
+      {:error, {:unknown_task_type, task_name, task_type}}
+    end
+  end
 
   defp parse_review(map, :review) do
     %Review{
--- a/core/lib/sykli/mcp/tools.ex
+++ b/core/lib/sykli/mcp/tools.ex
@@ -500,5 +500,11 @@ defmodule Sykli.MCP.Tools do
   defp format_error({:typescript_failed, msg}), do: "TypeScript SDK failed: #{msg}"
   defp format_error({:python_failed, msg}), do: "Python SDK failed: #{msg}"
   defp format_error({:missing_tool, tool, hint}), do: "Missing #{tool}: #{hint}"
+  defp format_error({:task_type_on_review, _} = reason), do: Sykli.Graph.format_error(reason)
+
+  defp format_error({:task_type_requires_version_3, _, _, _} = reason),
+    do: Sykli.Graph.format_error(reason)
+
+  defp format_error({:unknown_task_type, _, _} = reason), do: Sykli.Graph.format_error(reason)
   defp format_error(reason), do: inspect(reason)
 end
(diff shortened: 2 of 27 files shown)


yairfalse (commented):
## Overview

Phase 3B implementation: introduces `task_type` as the first version-3 semantic contract field. Closed 12-value enum, executable-task-only, rejected on review nodes. All five SDKs (Go, Rust, TypeScript, Python, Elixir) get explicit APIs in lockstep. Engine enforces version gating, enum validation, and review-rejection in both `parse_task_type/4` (graph.ex) and `check_task_types/3` (validate.ex). Schema gates `task_type` via top-level `if/then` on `version != "3"`. New conformance case `23-task-type.json` with fixtures in all five SDKs. Shared `Sykli.TaskType` module backs both engine paths.

## Strengths

- **Shared `Sykli.TaskType` module** (`core/lib/sykli/task_type.ex`) replaces the duplicated `@task_types ~w(...)` literal that existed in `graph.ex` and `validate.ex` on the previous branch. Both engine paths now call `Sykli.TaskType.valid?/1` / `Sykli.TaskType.all/0`. Single source of truth on the engine side.
- **All five SDKs updated in lockstep.** No "Go+Rust only" experimental caveat — Phase 3B doesn't repeat the regression PR #170 just retired for review nodes.
- **Conformance fixture `23-task-type.json` exists** with fixtures in all five SDKs. Cross-SDK byte-equality net is in place from day one.
- **Triple-layered enforcement is deliberate, not accidental.** Schema rejects task_type unless `version == "3"`; engine parser (`parse_task_type/4`) returns typed error tuples; engine validator (`check_task_types/3`) collects errors for the validate-CLI surface. The design doc explicitly asked whether engine should also enforce; the answer landed as "yes, both."
- **Engine error tuples have proper `format_error/1` clauses in both `graph.ex` and `validate.ex`.**
- **MCP error formatting delegates cleanly** (`mcp/tools.ex:503-509`) to `Sykli.Graph.format_error/1`. No duplicated rendering logic between MCP and engine.
- **Negative test coverage is real.** `core_test.exs` adds 5 tests covering: positive parse, version<3 rejection, unknown task_type, task_type-on-review, and format_error rendering. `validate_test.exs` mirrors the four error paths through the validate API. SDK suites cover invalid-value rejection and mixed v2/v3 emission.
- **TypeScript duck-typing protocol stayed clean.** `_hasTaskType()` is added to *both* `Task` and `Review` classes, preserving the structural invariant from PR #170 review issue #7.
- **Python uses idiomatic `Literal` + the existing `_validate_enum` helper** rather than rolling a new validator. Reuses the same pattern that already validates `criticality`, `on_fail`, `select_mode`, `gate_strategy`.
- **Auto-detection treats v3 as superset of v2 correctly.** Across all five SDKs, `has_v2_features` still drives `resources` block emission when version becomes "3". A pipeline with `task_type` *and* a container correctly emits version "3" with a populated `resources` block. Tested explicitly in 4/5 SDKs.
- **Docs updated comprehensively** — `## Versioning decision`, the version field bullet, the task object table, the known-gaps list, and the Phase 3 not-yet-included list. The "23 conformance fixtures" count was bumped. Doc grounding stays consistent.

## Issues

### Medium

**1. `format_error/1` prefix inconsistency between `graph.ex` and `validate.ex`.** Same logical error renders differently depending on which path surfaces it:

- `graph.ex:374-388`: `"Review node 'X' cannot declare task_type"` (no prefix)
- `validate.ex:301-303`: `"Error: Review node 'X' cannot declare task_type"` (with `Error: ` prefix)

MCP routes parse-time errors through `Sykli.Graph.format_error/1` (no prefix). The CLI `validate` path uses `Sykli.Validate.format_errors/1` which prepends `Error: `. So the same `task_type_on_review` mistake on the same pipeline reads as one of two different strings depending on which surface a user/agent hits.

This isn't a new inconsistency — `format_error/1` clauses for `:missing_command`, `:cycle`, `:missing_dependency`, etc. already use the `"Error: "` prefix in `validate.ex`. The new `task_type` clauses follow that pattern. The drift is between the *parse-time* errors (graph.ex, no prefix) and *validate-time* errors (validate.ex, prefix). Either:
- Decide once that all user-facing errors get `"Error: "` and update graph.ex's three new clauses, OR
- Decide that the prefix is validate.ex's responsibility and let graph.ex stay raw.

Either choice is fine. Mixing them is the bug.

**2. Intra-SDK list duplication is reducible.** The engine cleanly extracted `Sykli.TaskType`. Each SDK now has its own copies, which is necessary across language boundaries — but several SDKs duplicate the list *inside* their own codebase:

- **Elixir SDK (3 copies)**:
  - `sdk/elixir/lib/sykli/dsl.ex:122-135` — guard with atom literals
  - `sdk/elixir/lib/sykli/task.ex:64-75` — `@type task_type ::` spec
  - `sdk/elixir/lib/sykli/emitter.ex:7-21` — `@task_types` runtime list
- **Go SDK (2 copies)**: 12 `TaskType*` constants + `validTaskTypes map[TaskType]bool`. The map is derivable from the constants — a `[]TaskType` slice plus a function would work. Or: drop the map and use a `switch` in `TaskType()`.
- **Python SDK (2 copies)**: `TaskType = Literal[...]` (12 strings) + `TASK_TYPES: tuple[str, ...]` (12 strings). `typing.get_args(TaskType)` returns the Literal's values as a tuple at runtime — `TASK_TYPES = get_args(TaskType)` eliminates the duplicate.
- **TypeScript SDK (2 copies)**: `TaskType` union + `TASK_TYPES` array. Same constraint as Python (compile-time types vs runtime values), but `as const` arrays + derived union types (`type TaskType = (typeof TASK_TYPES)[number]`) collapse this.
- **Rust SDK (1 copy)**: enum variants + `as_str` match. Acceptable — match exhaustiveness gives compile-time safety.

Lowest-effort wins: Python (`get_args`) and TS (`as const` + derived union). Elixir's three copies are the highest payoff. Go's map elimination is purely cosmetic.

### Low / Nits

**3. Go SDK iteration discipline change.** The version-detection loop previously `break`-ed on first v2 detection (best case O(1)). The diff removes the `break` because it now also looks for v3 features:

```go
for _, t := range p.tasks {
    if t.taskType != "" {
        hasV3Features = true
    }
    if t.container != "" || len(t.mounts) > 0 {
        hasV2Features = true
    }
}
```

Now always O(N). For practical pipelines (~10-100 tasks), irrelevant — flagging the discipline change. A short-circuit `if hasV2Features && hasV3Features { break }` recovers the early exit at the cost of three lines.

**4. Empty-string sentinel in Python.** `Task._task_type: str = ""` uses `""` as "unset" rather than `Optional[str] = None`. Works because `_validate_enum` rejects `""`, so `if self._task_type:` is correct. Other SDKs use `nil` / `None` / `Option<TaskType>` / `undefined`. Stylistic.

**5. `__all__` exports `TaskType`** which is a `Literal` type alias, not a class. Users can use it for type hints (`def f(t: TaskType): ...`) but not `isinstance(x, TaskType)`. Standard Python typing trade-off; worth being aware of.

**6. Conformance fixture coverage is minimal.** Only two values exercised (`build`, `test`) in `23-task-type.json`. The other 10 enum values are exercised only by the schema validator (which checks the JSON Schema's enum) and by SDK unit tests. Conformance byte-equality across all 12 values would catch a hypothetical SDK that mis-stringifies an enum value (e.g., Rust's `as_str` typo). Not blocking.

**7. No mixed v2+v3 conformance fixture.** The case "task_type + container in the same pipeline emits version 3 with resources" is tested in unit suites for Go, Rust, Python, Elixir, TypeScript — but not in the cross-SDK conformance harness. A fixture with both a `task_type` and a `container` would lock byte equality on the corner case across all five SDKs.

**8. Conformance harness has no negative-case mechanism.** Schema validator (`scripts/validate-conformance-schema.py`) only checks positive emission against the schema. Negative scenarios (unknown task_type, task_type on review, version-mismatch) live in SDK unit tests + engine unit tests. The conformance model doesn't naturally accept negatives — fine, just naming the boundary.

## Risk Assessment

- **Engine-side risk: very low.** Three independent enforcement layers (schema `if/then`, parser, validator). Removing any one would still leave the other two.
- **Schema risk: low.** The `if: { properties: { version: { not: { const: "3" } } }, required: ["version"] }` rule depends on `version` being present, which the top-level `required: ["version", "tasks"]` already guarantees. If anyone ever loosens that top-level required, version-gating silently disables. Worth a one-line schema comment.
- **Cross-SDK divergence: low.** Conformance case 23 anchors byte equality for the most common positive case. Per-SDK unit tests cover invalid-value rejection.
- **Backward compatibility: clean.** task_type is opt-in. version "1" and "2" pipelines emit unchanged. Existing users see no wire-format difference until they call `.task_type(...)`.
- **Format-error inconsistency (issue 1) is user-visible but not behavioral.** No graph executes differently; just two surfaces render the same error two ways.

## Verdict

**Approve with one cleanup ask:**

1. Resolve the `format_error/1` prefix mismatch between `graph.ex` (no prefix) and `validate.ex` (`Error: ` prefix) — pick one and align the three new task_type clauses (issue 1).

**Optional follow-ups** (none blocking):

- Consolidate intra-SDK list duplications where the language allows (Python `get_args`, TS `as const` + derived union, Elixir shared module attribute, Go map elimination — issue 2).
- Add a mixed v2+v3 conformance fixture (issue 7).
- Expand conformance coverage across more enum values (issue 6).
- Add a one-line schema comment noting that the version gating depends on top-level `required: ["version"]` (risk-assessment item).

Strong PR. The previous reaction's three concrete asks (shared TaskType module, all five SDKs, conformance fixture) all landed exactly. The triple-layered enforcement is the right shape for a load-bearing schema-evolution event, and the discipline of doc → schema → engine → SDK lockstep is exactly what Phase 3 needed for its first slice.

yairfalse:
Addressed review feedback in 5ecf64f.

- Aligned parse-time task_type formatting with validate-time formatting by adding the `Error: ` prefix in `Sykli.Graph.format_error/1`; MCP still delegates to the graph formatter, so unrelated MCP error formatting remains unchanged.
- Reduced intra-SDK task_type enum duplication:
  - TypeScript now derives `TaskType` from the `TASK_TYPES as const` runtime list.
  - Python now derives `TASK_TYPES` from `typing.get_args(TaskType)` and uses `None` instead of an empty-string sentinel.
  - Elixir SDK now centralizes runtime task_type values in `Sykli.Task` and reuses that from DSL/emitter validation.
  - Go replaced the duplicate validation map with a switch over the exported constants and restored the version-detection early exit.

Validation rerun:
- schema JSON load passed
- `python3 scripts/validate-conformance-schema.py` -> 23 passed, 0 failed
- full conformance after Elixir warmup -> 140 passed, 0 failed, 0 skipped
- core `mix test` -> 1476 tests, 0 failures
- Go SDK tests passed
- TypeScript `vitest` and `tsc --noEmit` passed
- Python pytest -> 202 passed
- Elixir SDK tests -> 47 passed
- `git diff --check` passed

I left the broader optional conformance-expansion ideas as follow-ups rather than expanding this PR's contract surface.

PR #173, opened 2026-05-06
feat: add success_criteria metadata contract

yairfalse:
## Summary

Adds success_criteria as the Phase 3C-1 metadata contract field.

- adds success_criteria to the canonical schema as a version 3 executable-task-only field
- supports exit_code, file_exists, and file_non_empty criteria
- rejects success_criteria on review nodes, under versions 1/2, unknown criterion types, missing required criterion fields, and duplicate exit_code criteria
- adds core parse/validation support without executor enforcement
- adds explicit SDK APIs across Go, Rust, TypeScript, Python, and Elixir
- adds conformance case 24-success-criteria across all five SDKs
- documents that 3C-1 is metadata only and engine checking belongs in 3C-2 at the target level

## Non-goals

- no executor/result/retry/cache/target/runtime behavior changes
- no runtime enforcement yet
- no target replacement
- no capabilities/effects or other Phase 3 fields
- no preset inference

## Validation

- python3 scripts/validate-conformance-schema.py: 24 passed, 0 failed
- negative schema probe for version 2, review node, unknown type, missing fields, duplicate exit_code
- tests/conformance/run.sh: 145 passed, 0 failed, 0 skipped
- core mix test: 1483 tests, 0 failures, 1 skipped
- sdk/go go test ./...
- sdk/rust cargo test --lib
- sdk/typescript npx vitest run
- sdk/elixir mix test
- sdk/python PYTHONPATH=src python3.14 -m pytest: 206 passed

Note: the conformance harness still skips its embedded schema-validation step under python3.14 because that interpreter lacks jsonschema; standalone schema validation passed with local python3.


yairfalse (commented):
## Overview

Phase 3C-1 implementation: introduces `success_criteria` as a v3, executable-task-only metadata field. Three initial criterion types (`exit_code`, `file_exists`, `file_non_empty`), conjunctive (AND) semantics, at-most-one `exit_code` per task. New shared module `Sykli.SuccessCriteria` for engine parse + validate paths. All five SDKs gain explicit APIs. Conformance case `24-success-criteria.json` plus per-SDK fixtures. Schema rejects the field outside v3, on review nodes, with unknown types, on missing required keys, and on duplicate `exit_code`. **Metadata only — Phase 3C-1 explicitly defers executor enforcement to 3C-2.**

## Strengths

- **Shared `Sykli.SuccessCriteria` module** follows the `Sykli.TaskType` precedent set in PR #172. Single engine source of truth for the validation rules; both `graph.ex` parse path and `validate.ex` check path call into it.
- **Triple-layered enforcement** consistent with Phase 3B: schema `if/then`, parser (`graph.ex`), validator (`validate.ex`). Plus per-criterion `additionalProperties: false` + `oneOf` discrimination + `contains` + `maxContains: 1` for the exit_code-uniqueness constraint at the schema level. Five validation rules each enforced at two layers.
- **All five SDKs** with explicit APIs. No "Go/Rust only" caveat creeping back in.
- **Auto-detection** correctly bumps version to `"3"` when `success_criteria` is non-empty, with v3-as-superset-of-v2 preserved across all five SDKs.
- **Real negative-test coverage** in core: `validate_test.exs` exercises all 5 error paths (version<3, on review, unknown type, missing fields, duplicate exit_code). `core_test.exs` adds parse-path coverage. SDK-side: Rust has a `#[should_panic]` test for duplicate exit_code; Go and Elixir have analogous tests.
- **Format_error consistency restored** — the `graph.ex` no-prefix vs. `validate.ex` `Error:`-prefix inconsistency I flagged in PR #172 review was resolved upstream. `graph.ex`'s task_type clauses now consistently prefix `"Error: "`. New `success_criteria` clauses correctly follow the now-consistent pattern.
- **Engine restraint is documented and held.** The doc explicitly says "Phase 3C-1 stores, validates, and emits `success_criteria` only. It does not wire criteria into executor result, retry, cache, target, or runtime paths. The engine must not silently pretend to enforce criteria before target-level checking exists." The code matches the doc.
- **Type-safe representations where the language allows it.** Rust's `enum SuccessCriterion { ExitCode(i32), FileExists(String), FileNonEmpty(String) }` and TypeScript's discriminated union `{ type: 'exit_code'; equals: number } | …` make invalid combinations *unrepresentable* at compile time. Go's struct + constructor-function pattern (`ExitCode(0)`, `FileExists(path)`) achieves equivalent encapsulation. Elixir's DSL builders (`exit_code/1`, etc.) match.

## Issues

### High

**1. Whitespace pollution in `schemas/sykli-pipeline.schema.json`.** The diff shows tab characters introduced in 36 lines that the rest of the file indents with 2 spaces. Verified by grep — 36 tab-prefixed lines on the PR branch vs. 0 on main. JSON parses fine but the file is now mixed-indented and will produce huge diff churn the next time anything in the schema is touched.

```
                 { "required": ["k8s"] },
-                { "required": ["mounts"] },
-                { "required": ["timeout"] },
-                { "required": ["task_type"] }
-              ]
-            }
+	                { "required": ["mounts"] },
+	                { "required": ["timeout"] },
+	                { "required": ["task_type"] },
+	                { "required": ["success_criteria"] }
+	              ]
+	            }
```

The new content (the `success_criteria` line, the new `successCriterion`/`exitCodeCriterion`/etc. defs, the new `success_criteria` property) all use tab indentation. Looks like an editor auto-conversion. **Fix**: re-format the file with a JSON formatter using 2-space indent. `prettier --tab-width 2 --use-tabs false` is the quickest.

### Medium

**2. Python SDK has no constructor helpers — ergonomic regression vs. all other SDKs.** Python defines:

```python
SuccessCriterion = dict[str, Any]
```

…and exposes nothing else. Compare:

| SDK | User writes |
|---|---|
| Rust | `SuccessCriterion::ExitCode(0), SuccessCriterion::FileExists("coverage.out".into())` |
| Go | `sykli.ExitCode(0), sykli.FileExists("coverage.out")` |
| Elixir | `exit_code(0), file_exists("coverage.out")` |
| TS | `{ type: 'exit_code', equals: 0 }, { type: 'file_exists', path: 'coverage.out' }` (compile-time-checked union) |
| **Python** | `{"type": "exit_code", "equals": 0}, {"type": "file_exists", "path": "coverage.out"}` (raw dicts, validated only at `success_criteria(...)` call time) |

A Python user who writes `{"type": "file_existss", "path": "x.out"}` (one s too many) only finds out at SDK-validation time. The TypeScript user writing the equivalent gets caught at compile time. The Go/Rust/Elixir users *cannot* write that — the only way to construct a criterion is via the typed constructor.

The design doc names this Python shape as "likely" in `agent-contract-semantics.md`, so it's not against spec — but the doc says "likely SDK shapes," not "the Python SDK shall not provide constructors." The cheap fix is three module-level functions next to `_validate_success_criteria`:

```python
def exit_code(equals: int) -> SuccessCriterion:
    return {"type": "exit_code", "equals": equals}

def file_exists(path: str) -> SuccessCriterion:
    return {"type": "file_exists", "path": path}

def file_non_empty(path: str) -> SuccessCriterion:
    return {"type": "file_non_empty", "path": path}
```

…added to `__all__`, used in the conformance fixture (`success_criteria([exit_code(0), file_exists("coverage.out")])` instead of raw dicts). Closes the gap, costs ~10 lines. The dict shape stays accepted for users who want it.

**3. Validation message duplication between `Sykli.SuccessCriteria.format_error/1` and `validate.ex`'s `success_criteria_error_to_map/1`.** Both build user-facing strings from the same error tuples; the wording is *almost* identical but maintained in two places. Example: `Sykli.SuccessCriteria.format_error({:invalid_success_criteria, name, r})` returns `"Error: Task 'name' declares invalid success_criteria: r"`; `success_criteria_error_to_map/1` builds a `%{message: "Task 'name' declares invalid success_criteria: r"}` map that `validate.ex`'s `format_error/1` then prefixes with `"Error: "`. Same final string, two source-of-truth locations.

This is the same anti-pattern that exists for `task_type` (which has its own duplicate inline messages in `check_task_types/3`). Following precedent, but worth noting that Phase 3D (`capabilities`/`effects`) will introduce a third copy unless the validate path starts calling `Sykli.SuccessCriteria.format_error/1` directly to build its `:message` field.

**4. Elixir DSL validates criteria before checking the review-block guard.** `success_criteria(criteria)` first calls `validate_success_criteria!(criteria)`, then calls `update_current_task` which calls `reject_review_option!`. Inside a review block, an invalid criterion surfaces as `"invalid success_criteria type"` rather than `"success_criteria cannot be used inside a review block"`. The structural error should win; swap the order:

```elixir
def success_criteria(criteria) when is_list(criteria) do
  update_current_task(fn t ->
    reject_review_option!(t, "success_criteria")  # structural check first
    validate_success_criteria!(criteria)           # then content check
    %{t | success_criteria: t.success_criteria ++ criteria}
  end)
end
```

Minor UX nit.

### Low / Nits

**5. `Sykli.SuccessCriteria.normalize/1` is a post-validation passthrough.** By the time `normalize/1` runs, `validate/4` has already enforced `validate_no_extra_keys`. The explicit reconstruction `%{"type" => "exit_code", "equals" => equals}` is defensive but redundant — the input map already has only the allowed keys. Either drop the function (passthrough), or document why it exists (e.g., to canonicalize the key set in case future callers skip validate). Cosmetic.

**6. No `exit_code` range constraint.** Schema and engine accept any integer for `equals`. POSIX exit codes are 0-255; negative values or >255 are unsatisfiable in practice. Optional: add `minimum: 0, maximum: 255` to the `exitCodeCriterion` subschema. Phase 3C-1 is metadata-only so this doesn't break anything, but it's a cheap correctness improvement before 3C-2 wires in evaluation.

**7. Rust's `to_json` re-asserts empty paths** that `success_criteria(...)` already checked. Defensive duplication — fine, but the assertion fires at emit time rather than build time, which is the wrong layer if a user constructs the enum manually. The setter's check is the meaningful one; the `to_json` assertions are belt over belt.

**8. Empty `success_criteria: []` is accepted but doesn't trigger v3.** Go's `if len(t.successCriteria) > 0 { hasV3Features = true }`. The schema's `if/then` rejects `success_criteria` outside v3, but only when *required* — `required: ["success_criteria"]` evaluates true iff the key is present. So an explicit empty array in v1/v2 JSON would fail validation; SDKs avoid emitting empty arrays. Edge case is defended, just not tested.

## Risk Assessment

- **Engine-side risk: zero.** Phase 3C-1 is metadata-only by explicit design. Executor, retry, cache, target, runtime — all untouched.
- **Schema risk: low.** The `if/then` rule depends on top-level `required: ["version"]` (which is in place). Per-criterion subschemas use `additionalProperties: false`, catching typos. The `contains` + `maxContains: 1` exit_code uniqueness constraint is JSON Schema 2020-12, which is the schema's declared `$schema`.
- **Cross-SDK byte-level divergence: low.** Conformance case 24 anchors equality across all 5 SDKs.
- **Cross-SDK ergonomic divergence: medium** — see issue 2. Python users have a notably worse authoring experience than the other four.
- **Backward compatibility: clean.** Field is opt-in; v1/v2 pipelines unchanged; v3 pipelines without `success_criteria` unchanged.
- **Phase 3C-2 promises (engine evaluation) are clearly out of scope** — both the doc and the code make this explicit. Reduces risk that 3C-1 ships with half-implemented executor wiring.

## Verdict

1. **Re-indent the schema file** to remove the 36 tab-prefixed lines (issue 1). Pure style fix; mechanical.
2. **Add three Python constructor helpers** (`exit_code`, `file_exists`, `file_non_empty`) for SDK parity (issue 2). ~10 lines plus an `__all__` entry plus updating the Python conformance fixture and one or two Python tests to use them.

- Swap order of `reject_review_option!` and `validate_success_criteria!` in Elixir DSL (issue 4).
- Address the `format_error` duplication between `Sykli.SuccessCriteria` and `validate.ex` either now or as a Phase 3D-prep refactor (issue 3).
- Drop or document `Sykli.SuccessCriteria.normalize/1` (issue 5).
- Consider `minimum: 0, maximum: 255` on `exitCodeCriterion.equals` (issue 6).

Strong PR overall. The discipline of "metadata-only in 3C-1, evaluation in 3C-2" is exactly the right phasing for a feature that touches the executor's success/failure contract — landing schema and SDK plumbing without touching the result path lets the contract bake before semantics get wired in. Engine-side restraint is documented and held. The triple-layered enforcement (schema + parser + validator) follows the precedent from Phase 3B and continues to pay off.

yairfalse:
Addressed the review feedback in b6354e9.

Changes made:
- Reflowed schemas/sykli-pipeline.schema.json with 2-space JSON formatting; no tab indentation remains.
- Added Python helpers: exit_code(), file_exists(), file_non_empty(), and updated the Python conformance fixture/tests to use them.
- Centralized success_criteria validation wording through Sykli.SuccessCriteria.message/1 so parser/validator formatting stays aligned.
- Changed the Elixir DSL so review-block rejection wins before criterion-shape validation.
- Documented Sykli.SuccessCriteria.normalize/1 instead of leaving it unexplained.
- Added exit_code range validation as 0..255 across schema, core, SDK validation, docs, and tests.
- Added coverage for version 2 + empty success_criteria and out-of-range exit_code.

Validation run:
- rg -n "\t" schemas/sykli-pipeline.schema.json
- git diff --check
- python3 scripts/validate-conformance-schema.py -> 24 passed, 0 failed
- targeted schema negative probe for v2 empty success_criteria and exit_code range
- mix test test/sykli/graph/ai_native_test.exs test/sykli/validate_test.exs -> 38 tests, 0 failures
- cd sdk/go && env GOCACHE=/private/tmp/sykli-go-build go test ./...
- cd sdk/rust && cargo test --lib
- cd sdk/typescript && npx vitest run
- cd sdk/elixir && mix test
- cd sdk/python && env PYTHONPATH=src python3.14 -m pytest -> 208 passed
- env GOCACHE=/private/tmp/sykli-go-build SYKLI_CONFORMANCE_PYTHON=python3.14 tests/conformance/run.sh -> 145 passed, 0 failed, 0 skipped

Note: npx prettier could not be used because the sandboxed environment could not reach npm; I used python3 -m json.tool --indent 2 for the schema formatting instead. The conformance harness still reports the known embedded schema-validation skip under python3.14 because that interpreter lacks jsonschema, but standalone schema validation passed.

PR #175, opened 2026-05-07
Enforce contract schema version semantics in core engine

yairfalse:
## Summary

This PR makes Sykli enforce contract schema version semantics in the core engine. Contracts must now declare one of the repository supported pipeline wire-format/schema versions: `"1"`, `"2"`, or `"3"`. Missing, empty, malformed, or unsupported versions are rejected with explicit structured errors.

Note: the task prompt mentioned `"0.1"` as a possible current canonical version. This repository already defines the canonical contract versions as `"1"`, `"2"`, and `"3"` in the schema, SDKs, conformance fixtures, and docs, so this PR preserves and enforces that existing contract instead of introducing a new `"0.1"` version.

## Why

Sykli is a contract layer. The engine must not ignore the contract version. Without explicit version enforcement, SDKs and agents can emit contracts the engine does not actually understand, and future versions could be silently treated as older documents.

## What changed

- Added `Sykli.ContractSchemaVersion` as the single core policy for supported versions, current version, validation, and error formatting.
- Enforced version validation in `Sykli.Graph.parse/1` before task parsing.
- Enforced version validation in `Sykli.Validate.validate_json/1` with structured validation errors.
- Removed the previous silent default-to-`"1"` behavior for missing versions.
- Added positive core coverage for supported versions `"1"`, `"2"`, and `"3"`.
- Added negative core coverage for missing, null, empty string, numeric, boolean, array, object, unsupported future, unsupported major, and arbitrary string versions.
- Added schema-negative fixtures under `tests/conformance/schema-invalid/` for malformed and unsupported versions.
- Updated `scripts/validate-conformance-schema.py` to validate positive fixtures and assert negative schema fixtures are rejected.
- Updated `tests/conformance/run.sh` to check that both expected and SDK-emitted JSON declare a supported schema version.
- Updated README/docs to state that unsupported versions are rejected and future versions require explicit engine support.

## Validation

- [x] Schema JSON valid: `python3 -c "import json; json.load(open("schemas/sykli-pipeline.schema.json"))"` passed.
- [x] Schema fixtures passed: `python3 scripts/validate-conformance-schema.py` -> `Summary: 35 passed, 0 failed` (24 valid fixtures + 11 expected schema rejections).
- [x] Focused core parser/validator tests passed: `cd core && mix test test/core_test.exs test/sykli/validate_test.exs` -> `78 tests, 0 failures`.
- [x] Focused AI-native parser regression passed: `cd core && mix test test/sykli/graph/ai_native_test.exs` -> `7 tests, 0 failures`.
- [x] Full conformance passed: `env GOCACHE=/private/tmp/sykli-go-build SYKLI_CONFORMANCE_PYTHON=python3.14 tests/conformance/run.sh` -> `145 passed, 0 failed, 0 skipped`.
- [x] Core tests passed: `cd core && mix test` -> `1 property, 1492 tests, 0 failures, 1 skipped (13 excluded)`.
- [x] Go SDK tests passed: `cd sdk/go && env GOCACHE=/private/tmp/sykli-go-build go test ./...` -> passed.
- [x] Rust SDK tests passed: `cd sdk/rust && cargo test --lib` -> `111 passed, 0 failed`.
- [x] TypeScript SDK tests passed: `cd sdk/typescript && npx vitest run` -> `84 passed`.
- [x] Elixir SDK tests passed: `cd sdk/elixir && mix test` -> `52 tests, 0 failures`.
- [x] Python SDK tests passed: `cd sdk/python && env PYTHONPATH=src python3.14 -m pytest` -> `208 passed`.
- [x] Diff hygiene passed: `git diff --check` passed.

## Known notes

The conformance harness still reports the known embedded schema-validation skip under `python3.14` because that interpreter lacks `jsonschema`:

```text
Schema Validation
⚠ skipped — jsonschema package not installed for python3.14
```

Standalone schema validation was run with local `python3` and passed (`35 passed, 0 failed`). Full SDK conformance still passed (`145 passed, 0 failed, 0 skipped`).

## Out of scope

- Target-level `success_criteria` enforcement is intentionally not included.
- Review/gate/verify semantic normalization is intentionally not included.
- No runtime behavior, target behavior, or SDK feature expansion was attempted.
- No new contract fields were added.
- No compatibility mode for unknown future versions was introduced.

## Next recommended PR

The next recommended PR is target-level `success_criteria` enforcement, where the executor orchestrates criteria evaluation but each target evaluates criteria in its own execution context.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -277,7 +277,7 @@ Some cases carry `expected_failure: true`, which marks them as known-broken cont
   - `Sykli.ContractSchemaVersion` — supported `version` values + missing/empty/wrong-type/unsupported error policy.
 
   When adding a new closed-enum field or shared-policy concept, follow this pattern. SDKs must carry their own copies (separate Mix projects — engine modules are unreachable from `sdk/<lang>/`).
-- **Engine error formatting** — parse-time errors render via `Sykli.Graph.format_error/1` (also delegated to from `Sykli.MCP.Tools`); validate-time errors render via `Sykli.Validate.format_errors/1`. Validate-path strings prefix `"Error: "`; parse-path strings don't. Keep new error tuples consistent with the path that produces them.
+- **Engine error formatting** — parse-time errors render via `Sykli.Graph.format_error/1` (also delegated to from `Sykli.MCP.Tools`); validate-time errors render via `Sykli.Validate.format_errors/1`. Both paths prefix rendered strings with `"Error: "`. Keep new error tuples consistent with the path that produces them.
 
 ## CLI output rules
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -4,10 +4,12 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 ## Recent changes
 
-- **Pipeline contract is canonical.** `schemas/sykli-pipeline.schema.json` (JSON Schema 2020-12) is the source of truth for SDK-emitted JSON. `scripts/validate-conformance-schema.py` runs at the top of `tests/conformance/run.sh` and validates every fixture against it. Engine remains permissive; SDKs must emit only the canonical shape.
+- **Engine now enforces `version` strictly.** `Sykli.ContractSchemaVersion` (`core/lib/sykli/contract_schema_version.ex`) is the central policy module — it pins supported versions (`"1"`, `"2"`, `"3"`), the current version (`"3"`), and rejects missing/empty/wrong-type/unsupported versions. `Sykli.Graph.parse/1` and `Sykli.Validate.validate_data/1` both call `ContractSchemaVersion.fetch/1`. The previous silent default-to-`"1"` behavior is gone — payloads without a valid version are rejected. Negative coverage lives in `tests/conformance/schema-invalid/` (11 fixtures: missing, empty string, arbitrary string, integer, float, array, boolean, null, object, unsupported-future, unsupported-major).
+- **Phase 3C-1 `success_criteria`.** Three criterion types (`exit_code`, `file_exists`, `file_non_empty`), conjunctive AND, at-most-one `exit_code` per task. Engine vocabulary lives in `Sykli.SuccessCriteria`. **Metadata-only** — the executor does not evaluate criteria yet (deferred to 3C-2). All five SDKs ship explicit APIs. Conformance case `24-success-criteria.json`.
+- **Pipeline contract is canonical.** `schemas/sykli-pipeline.schema.json` (JSON Schema 2020-12) is the source of truth for SDK-emitted JSON. `scripts/validate-conformance-schema.py` validates positive fixtures in `tests/conformance/cases/` AND verifies negative fixtures in `tests/conformance/schema-invalid/` are rejected. SDKs must emit only the canonical shape.
 - **Phase 3B `task_type` (version `"3"`).** Closed 12-value enum (`build`, `test`, `lint`, `format`, `scan`, `package`, `publish`, `deploy`, `migrate`, `generate`, `verify`, `cleanup`) classifying executable tasks. Engine vocabulary lives in `Sykli.TaskType`. Triple-layered enforcement: schema `if/then`, `Sykli.Graph.parse_task_type/4`, `Sykli.Validate.check_task_types/3`. Rejected on review nodes. Design rationale: `docs/agent-contract-semantics.md`.
 - **`target` field removed from canonical SDK output.** All five SDK builder methods are deprecated no-ops (Python raises `DeprecationWarning`). The engine never read `target`; this was contract cleanup.
-- **Review nodes (`kind: "review"`) experimental** across all five SDKs. Schema rejects task-execution fields on reviews (`command`, `outputs`, `services`, `mounts`, `k8s`, `retry`, `timeout`, `task_type`).
+- **Review nodes (`kind: "review"`) experimental** across all five SDKs. Schema rejects task-execution fields on reviews (`command`, `outputs`, `services`, `mounts`, `k8s`, `retry`, `timeout`, `task_type`, `success_criteria`).
 - **GitHub-native foundation** shipped — GitHub App auth, webhook receiver (Plug + Bandit), Checks API client, `Sykli.Mesh.Roles`. See `docs/github-native.md`.
 - **CLI visual reset** shipped — Nordic-minimal renderer (`Sykli.CLI.Renderer/Theme/Live/FixRenderer`). The output rules are testable; banned vocabulary in §"CLI output rules" below.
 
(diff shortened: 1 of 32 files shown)


yairfalse (commented):
## Overview

Replaces the silent `Map.get(data, "version", "1")` default in both `Sykli.Graph.parse/1` and `Sykli.Validate.validate_data/1` with strict enforcement via a new `Sykli.ContractSchemaVersion` policy module. Missing, empty, malformed, or unsupported versions now produce structured errors. Adds 11 negative fixtures in `tests/conformance/schema-invalid/`, the schema validator now asserts each negative is rejected, and the conformance runner cross-checks that both expected and emitted JSON declare a supported version. Six existing test files updated to add `"version": "1"` to hand-rolled JSON. **Breaking change for any caller hand-rolling JSON without `version`.**

## Strengths

- **Quadruple-layered enforcement** for the version field: schema `enum` (rejects non-`"1"/"2"/"3"` values), `Sykli.ContractSchemaVersion.fetch/1` (parse path), `check_contract_schema_version/2` (validate path), and now the conformance runner (`json_version` + `supported_schema_version` checks for both expected and SDK-emitted JSON). Each layer can independently catch the failure mode.
- **`Sykli.ContractSchemaVersion` follows the established pattern.** Sits cleanly alongside `Sykli.TaskType` and `Sykli.SuccessCriteria`: exposes `valid?`-shaped predicates (`fetch/1`, `validate/1`), supplies both `format_error/1` and `to_error_map/1` for parse-time and validate-time render paths. The `supported_versions_text/0` helper means the error message ("supported versions: 1, 2, 3") is derived from the `@supported_versions` list, not hardcoded — bumping the supported list updates the messages automatically.
- **Comprehensive negative coverage.** 10 engine unit-test cases (`null`, `""`, `1`, `0.1`, `true`, `[]`, `{}`, `"0.2"`, `"1.0"`, `"banana"`) parallel 11 schema-invalid fixtures (same matrix plus `version-missing.json`). Engine and schema reject identically; the validator asserts each fixture fails.
- **Test-suite hygiene.** Six pre-existing test files (`attestation_test`, `context_test`, `executor_ai_hooks_test`, `explain_test`, `occurrence_test`, `plan_test`, plus `core_test`) updated to add `"version": "1"` rather than silently absorbing the new strictness behind a default-rewrite. Catching the breaking change in tests rather than CI is the right discipline.
- **`type_name/1` switches `null/integer/number/boolean/array/object`** — JSON-typed names in error messages, not Elixir-typed names. Users reading the error don't need to know `is_binary` vs `is_list` to understand "expected string, got array."
- **Conformance runner now validates SDK output** — even if a future SDK regresses and emits `"4"` or `null` for `version`, the runner catches it before the byte-equality phase. Defends against drift across the whole SDK matrix, not just the engine.
- **Docs updated comprehensively.** `docs/sdk-schema.md` Versioning Decision section rewritten from "advisory and partially version-aware" to "version-aware" with the explicit rejection rules. The 5-step migration plan that became outdated by this PR was correctly *removed* rather than left as a dead-list pointing at finished work.

## Issues

### Medium

**1. Cascading errors when version is invalid.** `validate.ex` extracts `version` as `nil` when `ContractSchemaVersion.fetch/1` returns an error, then continues to run `check_task_types/3` and `check_success_criteria/3` against `nil`. Result: a payload missing `version` with three tasks declaring `task_type` surfaces **four** errors:

```
Error: missing contract schema version
Error: Task 'a' declares task_type but pipeline version is not 3
Error: Task 'b' declares task_type but pipeline version is not 3
Error: Task 'c' declares task_type but pipeline version is not 3
```

The version error is the real problem; the three task_type errors are noise that disappears the moment the version error is fixed. Two ways to handle:

- **Short-circuit**: when `version_result` is `{:error, _}`, skip `check_task_types` and `check_success_criteria` (they're version-gated by definition; reporting them on a nil version is misleading).
- **Or document as intentional** — surface every problem so users fix everything in one pass.

I'd lean short-circuit. The cascading errors don't add information; they multiply noise proportional to v3-feature usage.

**2. CLAUDE.md "Engine error formatting" bullet is now factually wrong on this branch.** The existing bullet says:

> "Validate-path strings prefix `\"Error: \"`; parse-path strings don't. Keep new error tuples consistent with the path that produces them."

This was accurate at PR #172 time but isn't anymore: as of this branch, *all* parse-path `format_error/1` clauses in `graph.ex` (task_type × 3, success_criteria × 5, contract_schema_version × 4) return strings prefixed with `"Error: "`. The PR doesn't update the bullet, so the doc keeps telling future Claudes that parse-path strings are unprefixed. **Suggested fix** (bundle into this PR or a follow-up):

> "Both parse-time errors (rendered via `Sykli.Graph.format_error/1`, also reused by `Sykli.MCP.Tools`) and validate-time errors (rendered via `Sykli.Validate.format_errors/1`) prefix strings with `\"Error: \"`. Keep new error tuples consistent."

### Low / Nits

**3. Awkward version extraction in `validate.ex`.**

```elixir
version_result = Sykli.ContractSchemaVersion.fetch(data)
version = if match?({:ok, _}, version_result), do: elem(version_result, 1), else: nil
```

`match?` + `elem` is doing what a direct pattern-match expresses more naturally:

```elixir
version =
  case version_result do
    {:ok, v} -> v
    {:error, _} -> nil
  end
```

Same shape, easier to read. Even tighter would be inlining `version_result` into both the `check_contract_schema_version` call and the version extraction, but the explicit binding is fine for clarity.

**4. No whitespace-only version fixture or unit test.** `Sykli.ContractSchemaVersion.validate/1` does `String.trim(version) == ""` to catch `"   "` as empty, but no fixture or test exercises it. The schema would catch it via `enum` mismatch (different error code at the schema level: "value not in enum" vs the engine's `:empty_contract_schema_version`). Adding `tests/conformance/schema-invalid/version-whitespace-only.json` and a corresponding unit-test case would tighten coverage. Cheap.

**5. Error names are verbose.** `:missing_contract_schema_version`, `:unsupported_contract_schema_version`, etc. — six syllables when `:missing_version` / `:unsupported_version` would do. The verbosity probably guards against future ambiguity (a task could have its own "version" concept), but the contract field is *just* `"version"`, so the disambiguation isn't load-bearing yet. Style.

**6. Schema description still says version is "Pipeline wire-format/schema version. SDKs auto-detect: '3' if any executable task uses semantic fields…"** — the description doesn't mention the strict-rejection contract that this PR enforces. The `enum` constraint enforces it implicitly, but a downstream consumer reading the schema description doesn't know that missing `version` is now a validation error rather than a default-to-`"1"` fallback. Tiny tweak: add a sentence like *"Required; missing or unsupported values are rejected."* to the description.

## Risk Assessment

- **Breaking change scope: limited but real.** Any caller hand-rolling JSON without `version` will now fail to parse where it previously succeeded silently. The PR updated all six in-repo test files that did this. External callers (SDK output, MCP tooling, fixtures) all already emit `version`. The blast radius is contained, but anything in `eval/` or downstream consumers using hand-crafted JSON should be checked.
- **Schema risk: very low.** Schema's `enum` already rejected non-`"1"/"2"/"3"` values; this PR ensures the engine matches. Triple-layer enforcement (schema + parse + validate) means removing any one would still leave two.
- **Cross-SDK divergence: zero.** All five SDKs already emit `version` correctly via auto-detection. No SDK changes needed.
- **Conformance runner regression risk: low.** The new `json_version` / `supported_schema_version` checks catch the "SDK emits unsupported version" failure mode early and clearly.
- **Backward compatibility for stored JSON:** `.sykli/occurrence.json` and similar persisted artifacts that include task graphs may need a one-time migration if they were ever written without `version`. Worth verifying that nothing in the persistence path emits versionless task graphs. (Not addressed in PR — likely fine since SDKs always emit version, but worth a grep.)

## Verdict

1. **Short-circuit downstream version-gated checks when version is invalid** (issue 1) — return early after `check_contract_schema_version` or pass the version-result through downstream checks so they skip on `:error`.
2. **Update the CLAUDE.md "Engine error formatting" bullet** (issue 2) — it now contradicts the parse-path clauses on this branch. Either fix the bullet or fix any clause that's still inconsistent.

- Inline-pattern-match version extraction in `validate.ex` (issue 3).
- Add a whitespace-only version fixture + test (issue 4).
- Tighten the schema's `version` description to mention strict rejection (issue 6).

Strong PR. The triple-layer enforcement is the right shape for a load-bearing schema-evolution event, and the discipline of *updating* every test that hand-rolled versionless JSON (rather than silently absorbing the breaking change) is exactly how to ship a contract tightening cleanly. The new `Sykli.ContractSchemaVersion` module slots into the established `Sykli.TaskType` / `Sykli.SuccessCriteria` pattern, and the conformance-runner cross-check on emitted version is a defense-in-depth move that pays off the moment any future SDK regresses.

PR #176, opened 2026-05-07
Add target-level success_criteria enforcement

yairfalse:
Summary
-------
This PR makes success_criteria enforceable at the target boundary. The executor now orchestrates criteria evaluation after a task command succeeds, while targets evaluate criteria in the execution context they own. Local target support is implemented first, and unsupported target/runtime combinations fail explicitly instead of silently skipping criteria.

Why
---
Sykli is a contract layer. A task cannot be considered successful only because its command exited 0 if the contract declares additional success criteria. The engine must verify that the execution satisfied the contract, or fail clearly when the active target cannot evaluate the criteria.

What changed
------------
- Added a target-level success_criteria evaluation callback.
- Added structured success criterion result objects.
- Implemented local target evaluation for exit_code, file_exists, and file_non_empty.
- Made failed criteria fail the task/run after command success.
- Made unsupported criteria/targets fail explicitly with unsupported_success_criteria_for_target.
- Bypassed cache hits for tasks with success_criteria so cached success does not pretend to prove target-owned checks.
- Added success_criteria result visibility on executor task results and CLI JSON result maps.
- Added focused tests for passing, failing, unsupported, retry, command-failure, no-criteria, path-safety, and workdir behavior.
- Updated docs to describe target-level enforcement semantics.

Validation
----------
- [x] Schema JSON load passed: python3 -c "import json; json.load(open('schemas/sykli-pipeline.schema.json'))"
- [x] Schema fixtures passed: python3 scripts/validate-conformance-schema.py -> 36 passed, 0 failed
- [x] Focused core tests passed: mix test test/sykli/executor/success_criteria_enforcement_test.exs -> 14 tests, 0 failures
- [x] Focused parser/schema tests passed: mix test test/sykli/executor/success_criteria_enforcement_test.exs test/sykli/validate_test.exs test/sykli/graph/ai_native_test.exs -> 55 tests, 0 failures
- [x] Full core suite passed: mix test -> 1 property, 1507 tests, 0 failures, 1 skipped (13 excluded)
- [x] Full conformance passed: env GOCACHE=/private/tmp/sykli-go-build SYKLI_CONFORMANCE_PYTHON=python3.14 tests/conformance/run.sh -> 145 passed, 0 failed, 0 skipped
- [x] Go SDK tests passed: env GOCACHE=/private/tmp/sykli-go-build go test ./...
- [x] Rust SDK tests passed: cargo test --lib -> 111 passed, 0 failed
- [x] TypeScript SDK tests passed: npx vitest run -> 84 passed
- [x] Elixir SDK tests passed: mix test -> 52 tests, 0 failures
- [x] Python SDK tests passed: env PYTHONPATH=src python3.14 -m pytest -> 208 passed
- [x] git diff --check passed

Known notes
-----------
- The conformance harness still reports the known embedded schema-validation skip under python3.14 because that interpreter lacks jsonschema. Standalone schema validation passes with local python3.
- A direct Go test run attempted to use the default Go cache under ~/Library/Caches and hit sandbox permissions. Re-running with GOCACHE=/private/tmp/sykli-go-build passed.
- A direct Python SDK pytest run without PYTHONPATH could not import the local src package. Re-running with PYTHONPATH=src passed.

Out of scope
------------
- Docker/Podman/Kubernetes criteria evaluation is intentionally not implemented.
- Review/gate/verify semantic normalization is intentionally not included.
- Schema version behavior is unchanged.
- No broad runtime redesign was attempted.
- No UI/backend work was included.

Next recommended PR
-------------------
The next recommended PR is to add target-specific criteria support for Docker/Podman/Kubernetes using the same target-level interface.

### DIFF
--- a/core/lib/sykli/cache.ex
+++ b/core/lib/sykli/cache.ex
@@ -172,6 +172,10 @@ defmodule Sykli.Cache do
   def format_miss_reason(:corrupted), do: "cache corrupted"
   def format_miss_reason(:blobs_missing), do: "outputs missing"
   def format_miss_reason(:config_changed), do: "config changed"
+
+  def format_miss_reason(:success_criteria_requires_execution),
+    do: "success criteria require execution"
+
   def format_miss_reason(other), do: "#{other}"
 
   @doc """
--- a/core/lib/sykli/cli.ex
+++ b/core/lib/sykli/cli.ex
@@ -309,12 +309,14 @@ defmodule Sykli.CLI do
   end
 
   defp task_result_to_map(%Sykli.Executor.TaskResult{} = r) do
-    base = %{
-      name: r.name,
-      status: Atom.to_string(r.status),
-      duration_ms: r.duration_ms,
-      cached: r.status == :cached
-    }
+    base =
+      %{
+        name: r.name,
+        status: Atom.to_string(r.status),
+        duration_ms: r.duration_ms,
+        cached: r.status == :cached
+      }
+      |> maybe_add_success_criteria_results(r.success_criteria_results)
 
     case r.error do
       %Sykli.Error{} = err ->
@@ -326,6 +328,25 @@ defmodule Sykli.CLI do
       other ->
         Map.put(base, :error, %{code: "unknown", message: inspect(other), hints: []})
     end
+  end
+
+  defp maybe_add_success_criteria_results(base, []), do: base
+
+  defp maybe_add_success_criteria_results(base, results) do
+    Map.put(base, :success_criteria_results, Enum.map(results, &success_criteria_result_to_map/1))
+  end
+
+  defp success_criteria_result_to_map(result) do
+    %{
+      index: result.index,
+      type: result.type,
+      status: Atom.to_string(result.status),
+      message: result.message,
+      target: result.target,
+      evidence: result.evidence
+    }
+    |> Enum.reject(fn {_key, value} -> is_nil(value) end)
+    |> Map.new()
   end
 
   @doc false
--- a/core/lib/sykli/error.ex
+++ b/core/lib/sykli/error.ex
@@ -9,15 +9,15 @@ defmodule Sykli.Error do
   - Actionable hints for every error
   - Visual consistency across the CLI
 
-   ## Error Codes
+  ## Error Codes
 
-   | Code | Category | Description |
-   |------|----------|-------------|
-   | task_failed | execution | Task command failed |
-   | task_timeout | execution | Task timed out |
-   | success_criteria_failed | execution | Declared success criteria failed |
-   | unsupported_success_criteria_for_target | execution | Target cannot evaluate declared success criteria |
-   | missing_secrets | execution | Missing secrets |
+  | Code | Category | Description |
+  |------|----------|-------------|
+  | task_failed | execution | Task command failed |
+  | task_timeout | execution | Task timed out |
+  | success_criteria_failed | execution | Declared success criteria failed |
+  | unsupported_success_criteria_for_target | execution | Target cannot evaluate declared success criteria |
+  | missing_secrets | execution | Missing secrets |
   | dependency_cycle | validation | Circular dependency in task graph |
   | invalid_service | validation | Invalid service config |
   | invalid_mount | validation | Invalid mount config |
--- a/core/lib/sykli/error.ex
+++ b/core/lib/sykli/error.ex
@@ -9,13 +9,15 @@ defmodule Sykli.Error do
   - Actionable hints for every error
   - Visual consistency across the CLI
 
-  ## Error Codes
+   ## Error Codes
 
-  | Code | Category | Description |
-  |------|----------|-------------|
-  | task_failed | execution | Task command failed |
-  | task_timeout | execution | Task timed out |
-  | missing_secrets | execution | Missing secrets |
+   | Code | Category | Description |
+   |------|----------|-------------|
+   | task_failed | execution | Task command failed |
+   | task_timeout | execution | Task timed out |
+   | success_criteria_failed | execution | Declared success criteria failed |
+   | unsupported_success_criteria_for_target | execution | Target cannot evaluate declared success criteria |
+   | missing_secrets | execution | Missing secrets |
   | dependency_cycle | validation | Circular dependency in task graph |
   | invalid_service | validation | Invalid service config |
   | invalid_mount | validation | Invalid mount config |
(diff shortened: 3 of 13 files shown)


yairfalse (commented):
## Overview

Phase 3C-2 implementation: makes `success_criteria` enforceable at the target boundary. PR #173 landed the parse/schema/SDK plumbing as metadata-only and explicitly deferred executor wiring; this PR delivers it. The executor orchestrates criteria evaluation after task command success; targets evaluate in their own execution context. Local shell target supports `exit_code`, `file_exists`, `file_non_empty`. K8s, Mesh, and Local-with-container fail explicitly with `unsupported_success_criteria_for_target` rather than silently skipping. Cached tasks bypass cache when criteria are declared, since a cache hit cannot prove target-owned checks. New `Sykli.SuccessCriteria.Result` struct carries structured per-criterion results through `Executor.TaskResult` to the CLI/MCP surface. 14-test focused suite covers the happy paths, all failure modes, retry interaction, path-security, and unsupported-target paths.

## Strengths

- **Architecture is clean three-tier separation.** Target.Behaviour declares `@callback evaluate_success_criteria/4` with `@optional_callbacks` so existing targets aren't forced to implement. Executor calls the callback (or surfaces `unsupported_success_criteria_for_target` if the target doesn't implement). Result struct (`index`, `type`, `status`, `message`, `evidence`, `target`) carries structured data through to CLI JSON output via `task_result_to_map` / `success_criteria_result_to_map`.
- **Failure modes are loud, not quiet.** K8s and Mesh explicitly implement the callback to return `:unsupported`; Local explicitly returns `:unsupported` for the container-runtime case. There's no path where "the target can't check this" silently turns into "the task passed."
- **Cache bypass is correct.** `if success_criteria(task) == [] do CacheService.check_and_restore else {:miss, nil, :success_criteria_requires_execution}`. The new miss reason is added to `Sykli.Cache.format_miss_reason/1`. A cached result cannot prove that workdir-state-dependent criteria still hold.
- **Path security reuses the documented project pattern.** `resolve_criterion_path/2` rejects absolute paths and uses the existing `path_within?(path, base)` helper (`String.starts_with?(path, base <> "/")`) that's already declared in CLAUDE.md as the project-wide path-traversal guard.
- **Backward-compat run_task return shapes.** New `{:ok, output, criteria_results}` and `{:error, reason, criteria_results}` 3-tuples coexist with legacy 2-tuples. Targets that don't yet emit the 3-tuple still work; the executor handles both. The retry path is duplicated correctly for both shapes (with `_success_criteria_results` discarded on retry — intentional, since the next attempt re-evaluates).
- **Per-criterion evidence in results.** `%{expected: 0, actual: 0, path: "...", resolved_path: "...", size: 42}` etc. Downstream consumers (CLI, MCP, future TUI) get structured data rather than parsed strings.
- **Real test coverage** — 14 tests including: happy paths for all three criterion types, command-success-but-criterion-fail, command-failure-doesn't-evaluate-criteria, retry interaction (verifies attempts counter), unsupported target via mock, container-runtime unsupported, no-criteria preserves behavior, workdir-relative resolution, `../` escape rejection, absolute-path rejection. The `UnsupportedTarget` mock testing the `function_exported?` fallback path is the right shape.
- **Two new `Sykli.Error` constructors** (`success_criteria_failed/3`, `unsupported_success_criteria_for_target/4`) follow the existing pattern with hints, notes, and structured `criterion_notes/1` formatting that surfaces evidence inline.

## Issues

### Medium

**1. `File.stat/1` follows symlinks — symlink TOCTOU lets criteria be satisfied by files outside the workdir.** The path-traversal check operates on the *declared path string*, but `File.stat/1` follows symlinks. A task command that creates `coverage.out` as a symlink to `/etc/passwd` (or any file outside the workdir) lets a criterion declaring `path: "coverage.out"` be satisfied by the symlink target's stat. The literal path is within the workdir; the resolved file isn't.

Threat model is genuinely mild here — this is local execution and the task author writes both the command and the criterion — but the contract advertises "evaluated relative to task workdir," and that contract leaks. Two cleanups:

- Use `File.lstat/1` to detect symlinks. Either reject them outright (cleanest, matches the "no escape" promise), or follow them and re-validate the resolved target is `path_within?(workdir)`.
- Add a test: create `inside.txt` as a symlink to `/tmp/outside-target`, declare `file_exists path: "inside.txt"`, assert it fails with "path escapes task workdir" or similar.

This is the kind of contract-leak that's cheap to fix now and expensive to fix later when external tools start trusting the "within workdir" promise.

**2. Cache bypass is silent and total — performance regression risk for any task with `success_criteria`.** Adding *any* `success_criteria` to a task disables caching for that task entirely. The PR description names the trade-off ("Bypassed cache hits for tasks with success_criteria so cached success does not pretend to prove target-owned checks") but `docs/sdk-schema.md`'s `success_criteria` section doesn't surface it to users. Someone who added `success_criteria` to their hot test task during 3C-1 (when it was metadata-only) will see test wall-clock time jump on this release.

Two paths:
- **Document it now** in the `success_criteria` doc section: "Tasks with declared criteria do not currently use cache hits, since cached output cannot prove target-owned checks."
- **Optimize selectively in 3C-3** — `exit_code` is structurally provable from a cached `:passed` (the cache key already encodes that the cached run exited successfully). Only `file_exists` / `file_non_empty` genuinely need re-execution because workdir state can differ. A criterion-aware cache decision could preserve hits for exit_code-only tasks.

For this PR, just adding the doc note is sufficient. The selective bypass is a future optimization, not a correctness fix.

**3. `Sykli.Error` moduledoc table indent regression.** The diff at `error.ex:9-22` shifts the `## Error Codes` table from 2-leading-space indent to 3-leading-space indent for the lines this PR touches. The rows above and below stay at 2-space. Result: mixed indentation in a markdown table inside an Elixir `@moduledoc`. ExDoc renders heredoc literally; the indent shift breaks table rendering for the affected rows. Pure whitespace fix — re-indent the new rows to match the surrounding 2-space.

### Low / Nits

**4. `function_exported?(target, :evaluate_success_criteria, 4)` and `function_exported?(target, :name, 0)` are now dead code for all in-tree targets.** Local, K8s, and Mesh all explicitly implement both callbacks. The runtime checks are only meaningful for hypothetical 3rd-party targets or test mocks like `UnsupportedTarget`. Worth keeping as defense-in-depth; just noting that they're not load-bearing for in-tree behavior.

**5. Unknown-criterion-type message reads as "this target can't" rather than "this is malformed".** `criterion_unsupported(index, type, "unsupported success_criteria type #{inspect(type)}", ...)` produces the same `:unsupported` result that the container-can't-evaluate path produces. Schema validation should reject unknown types upstream of this code, so reaching this clause means schema-bypass — semantically more like `:failed` with "unrecognized type" than `:unsupported`. Status-wise either is defensible; the message wording could be `"unrecognized success_criteria type"` to distinguish from "target unsupported."

**6. `docs/error-codes.md` not updated** (referenced from CLAUDE.md as the canonical catalog of public error codes with stability tiers). The two new codes — `success_criteria_failed`, `unsupported_success_criteria_for_target` — only appear in `error.ex`'s moduledoc table. Adding them to `docs/error-codes.md` keeps the catalog complete for downstream tooling that depends on stable codes.

**7. `task_result_to_map` cleanly surfaces `success_criteria_results` only when non-empty.** Good — keeps the JSON envelope quiet for the common case (no criteria declared). The `success_criteria_result_to_map/1` helper rejects nil-valued keys before building the map, so the JSON shape is minimal. Nice.

## Risk Assessment

- **Engine surface change: significant but cleanly handled.** New optional Target.Behaviour callback. All in-tree targets implement it; 3rd-party targets degrade to the explicit `unsupported_success_criteria_for_target` error rather than silent failure.
- **Run-task return shape: extended.** 3-tuple shapes added; 2-tuple shapes still handled. Retry path covers both.
- **Cache hit-rate regression: real but bounded.** Tasks with criteria always re-execute. Quantitative impact depends on criteria adoption.
- **Path security: tested for the obvious cases (absolute, `..`); symlink case undocumented and untested** (issue 1).
- **Cross-target divergence: zero confusion.** Only Local supports evaluation; K8s, Mesh, and Local-with-container fail loudly.
- **Phase 3C-1 → 3C-2 staging discipline held.** The metadata-only 3C-1 did exactly what the design doc promised; 3C-2 lifts that caveat at the executor layer without touching SDKs or schema.

## Verdict

1. **Fix the moduledoc indent regression** in `core/lib/sykli/error.ex` (issue 3) — pure whitespace, mechanical.
2. **Address the symlink-follow case** (issue 1) — switch to `File.lstat/1` and either reject symlinks or re-validate the resolved target. Add a test.

- Document the cache-bypass behavior in `docs/sdk-schema.md`'s `success_criteria` section (issue 2).
- Update `docs/error-codes.md` with the two new public codes (issue 6).
- Reword the unknown-criterion-type message (issue 5).
- Future: selective cache bypass — only criteria that depend on workdir state need re-execution; `exit_code` could honor cache hits (issue 2 follow-up).

Strong PR. The Target.Behaviour optional-callback shape is the right way to land this — Local can ship enforcement without forcing K8s/Mesh implementations, and 3rd-party targets degrade cleanly. Failure-modes-are-loud discipline is exactly right for a contract enforcement PR; silent skipping would have been the trap. The 14-test focused suite covers most edge cases, with the symlink case being the only obvious gap.

PR #187, opened 2026-05-08
Add local work item state model

yairfalse:
## Summary

Adds the local Work Item model and file-backed store for Team Mode Phase 1. Work Items are the local coordination unit for agentic work before any coordinator or network sync exists.

This PR intentionally stops at the local state model. It does not add CLI commands, coordinator endpoints, daemon join, run association, gate state, MCP tools, or remote execution.

## Why Work Item exists

Team Mode needs a durable unit of coordination before it can sync team state. A Work Item represents work such as reviewing a PR, investigating a failing deployment, generating a patch, or checking API breakage. This PR makes that state real locally under `.sykli/` while preserving local-first behavior.

## Local file path

Work items are persisted at:

```text
.sykli/work/items/<id>.json
```

The store validates IDs before path construction and rejects traversal-like IDs such as `../escape` or absolute paths.

## JSON shape

```json
{
  "id": "01KQPF7Q4W6J2M7V6YF2N0H6A2",
  "version": "1",
  "title": "Review PR #176",
  "intent": "Check timeout and success criteria behavior",
  "status": "open",
  "created_by": "member:yair",
  "assigned_to_type": null,
  "assigned_to_id": null,
  "created_at": "2026-05-08T10:00:00Z",
  "updated_at": "2026-05-08T10:00:00Z",
  "notes": []
}
```

Allowed statuses:

- `open`
- `claimed`
- `running`
- `blocked`
- `done`
- `failed`
- `cancelled`

Allowed assignment types:

- `member`
- `agent`
- `daemon`
- `null` when unassigned

## What changed

- Added `Sykli.WorkItem` with constructors, JSON map conversion, status validation, assignment validation, claim support, and note append support.
- Added `Sykli.Work.Store` for `.sykli/work/items/<id>.json` persistence.
- Store writes are atomic via temp file + rename.
- Store returns explicit structured error tuples for missing items, malformed JSON, invalid IDs, invalid statuses, invalid assignment types, and write/read failures.
- Updated `docs/local-state-plane.md` with the local work item path, JSON shape, enum values, and privacy boundary.

## Validation

Ran:

- `cd core && mix format lib/sykli/work_item.ex lib/sykli/work/store.ex test/sykli/work_item_test.exs test/sykli/work/store_test.exs` → passed
- `git diff --check` → passed
- `cd core && mix test test/sykli/work_item_test.exs test/sykli/work/store_test.exs` → 21 tests, 0 failures
- `cd core && mix test` → 1 property, 1539 tests, 0 failures, 1 skipped, 13 excluded

## Oracle / blackbox status

No oracle or blackbox cases were added in this PR because there is no user-facing CLI surface yet. PR 2 adds `sykli work ...` commands and should add CLI, blackbox, and oracle coverage for create/list/show/claim/note JSON behavior.

## Out of scope

- CLI commands
- coordinator service
- daemon join / heartbeat
- team sync
- run association
- gate state
- MCP tools
- Kubernetes deployment
- remote execution

## Next PR

Next: **Add local work item CLI commands**

Suggested scope:

- `sykli work create`
- `sykli work list`
- `sykli work show`
- `sykli work claim`
- `sykli work note`
- `--json` output for every command
- CLI tests, blackbox workflow, and oracle JSON expectations

### DIFF
--- a/core/lib/sykli/work/store.ex
+++ b/core/lib/sykli/work/store.ex
@@ -5,6 +5,11 @@ defmodule Sykli.Work.Store do
   The store persists work items under `.sykli/work/items/<id>.json` in the
   requested project path. It is intentionally local-only; coordinator sync is a
   later Team Mode phase.
+
+  Writes are atomic per file, but this store does not provide cross-process
+  locking. Read-modify-write operations, including claims, enforce model rules
+  for the loaded item and remain last-writer-wins under concurrent writers until
+  the coordinator store adds transactional claims.
   """
 
   alias Sykli.WorkItem
@@ -120,9 +125,10 @@ defmodule Sykli.Work.Store do
   end
 
   defp encode(%WorkItem{} = item) do
-    {:ok, Jason.encode!(WorkItem.to_map(item), pretty: true)}
-  rescue
-    error -> {:error, {:work_item_encode_failed, error}}
+    case Jason.encode(WorkItem.to_map(item), pretty: true) do
+      {:ok, json} -> {:ok, json}
+      {:error, error} -> {:error, {:work_item_encode_failed, error}}
+    end
   end
 
   defp decode(json, path) do
--- a/core/lib/sykli/work/store.ex
+++ b/core/lib/sykli/work/store.ex
@@ -1,0 +1,159 @@
+defmodule Sykli.Work.Store do
+  @moduledoc """
+  File-backed local work item store.
+
+  The store persists work items under `.sykli/work/items/<id>.json` in the
+  requested project path. It is intentionally local-only; coordinator sync is a
+  later Team Mode phase.
+  """
+
+  alias Sykli.WorkItem
+
+  @items_dir ".sykli/work/items"
+
+  @doc "Returns the work item storage directory for a base path."
+  def items_dir(opts \\ []) do
+    opts
+    |> base_path()
+    |> Path.join(@items_dir)
+  end
+
+  @doc "Creates and persists a new work item."
+  def create(title, opts \\ []) do
+    with {:ok, item} <- WorkItem.new(title, opts),
+         :ok <- save(item, opts) do
+      {:ok, item}
+    end
+  end
+
+  @doc "Persists an existing work item."
+  def save(%WorkItem{} = item, opts \\ []) do
+    with :ok <- WorkItem.validate_id(item.id),
+         {:ok, path} <- item_path(item.id, opts),
+         :ok <- File.mkdir_p(Path.dirname(path)),
+         {:ok, json} <- encode(item),
+         :ok <- atomic_write(path, json) do
+      :ok
+    end
+  end
+
+  @doc "Loads a work item by id."
+  def get(id, opts \\ []) do
+    with {:ok, path} <- item_path(id, opts) do
+      case File.read(path) do
+        {:ok, json} -> decode(json, path)
+        {:error, :enoent} -> {:error, {:work_item_not_found, id}}
+        {:error, reason} -> {:error, {:work_item_read_failed, id, reason}}
+      end
+    end
+  end
+
+  @doc "Lists all persisted work items in deterministic id order."
+  def list(opts \\ []) do
+    dir = items_dir(opts)
+
+    case File.ls(dir) do
+      {:ok, files} ->
+        files
+        |> Enum.filter(&String.ends_with?(&1, ".json"))
+        |> Enum.sort()
+        |> Enum.reduce_while({:ok, []}, fn file, {:ok, items} ->
+          id = String.trim_trailing(file, ".json")
+
+          case get(id, opts) do
+            {:ok, item} -> {:cont, {:ok, [item | items]}}
+            {:error, reason} -> {:halt, {:error, reason}}
+          end
+        end)
+        |> case do
+          {:ok, items} -> {:ok, Enum.reverse(items)}
+          error -> error
+        end
+
+      {:error, :enoent} ->
+        {:ok, []}
+
+      {:error, reason} ->
+        {:error, {:work_item_list_failed, reason}}
+    end
+  end
+
+  @doc "Updates a persisted work item's status."
+  def update_status(id, status, opts \\ []) do
+    with {:ok, item} <- get(id, opts),
+         {:ok, updated} <- WorkItem.update_status(item, status, opts),
+         :ok <- save(updated, opts) do
+      {:ok, updated}
+    end
+  end
+
+  @doc "Claims a persisted work item."
+  def claim(id, assignment_type, assignment_id, opts \\ []) do
+    with {:ok, item} <- get(id, opts),
+         {:ok, updated} <- WorkItem.claim(item, assignment_type, assignment_id, opts),
+         :ok <- save(updated, opts) do
+      {:ok, updated}
+    end
+  end
+
+  @doc "Appends a note to a persisted work item."
+  def append_note(id, body, opts \\ []) do
+    with {:ok, item} <- get(id, opts),
+         {:ok, updated} <- WorkItem.append_note(item, body, opts),
+         :ok <- save(updated, opts) do
+      {:ok, List.last(updated.notes), updated}
+    end
+  end
+
+  defp item_path(id, opts) do
+    with :ok <- WorkItem.validate_id(id) do
+      dir = items_dir(opts)
+      path = Path.expand(Path.join(dir, "#{id}.json"))
+      expanded_dir = Path.expand(dir)
+
+      if path_within?(path, expanded_dir) do
+        {:ok, path}
+      else
+        {:error, {:work_item_path_escape, id}}
+      end
+    end
+  end
+
+  defp encode(%WorkItem{} = item) do
+    {:ok, Jason.encode!(WorkItem.to_map(item), pretty: true)}
+  rescue
+    error -> {:error, {:work_item_encode_failed, error}}
+  end
+
+  defp decode(json, path) do
+    with {:ok, data} <- Jason.decode(json),
+         {:ok, item} <- WorkItem.from_map(data) do
+      {:ok, item}
+    else
+      {:error, %Jason.DecodeError{} = error} -> {:error, {:malformed_work_item_json, path, error}}
+      {:error, reason} -> {:error, reason}
+    end
+  end
+
+  defp atomic_write(path, json) do
+    tmp_path = "#{path}.tmp-#{System.unique_integer([:positive])}"
+
+    with :ok <- File.write(tmp_path, json),
+         :ok <- File.rename(tmp_path, path) do
+      :ok
+    else
+      {:error, reason} ->
+        File.rm(tmp_path)
+        {:error, {:work_item_write_failed, path, reason}}
+    end
+  end
+
+  defp base_path(opts), do: Keyword.get(opts, :path, ".")
+
+  defp path_within?(path, base) do
+    expanded_path = Path.expand(path)
+    expanded_base = Path.expand(base)
+
+    expanded_path == expanded_base or String.starts_with?(expanded_path, expanded_base <> "/")
+  end
+end
--- a/core/lib/sykli/work_item.ex
+++ b/core/lib/sykli/work_item.ex
@@ -1,0 +1,241 @@
+defmodule Sykli.WorkItem do
+  @moduledoc """
+  Local work item model for Team Mode.
+
+  A work item is a local coordination record stored under
+  `.sykli/work/items/<id>.json`. Networking and coordinator sync are added in
+  later Team Mode phases; this module only defines the local shape and
+  validation rules.
+  """
+
+  @version "1"
+  @statuses ~w(open claimed running blocked done failed cancelled)
+  @assignment_types ~w(member agent daemon)
+
+  @enforce_keys [:id, :title, :status, :created_at, :updated_at]
+  defstruct [
+    :id,
+    version: @version,
+    title: nil,
+    intent: nil,
+    status: "open",
+    created_by: nil,
+    assigned_to_type: nil,
+    assigned_to_id: nil,
+    created_at: nil,
+    updated_at: nil,
+    notes: []
+  ]
+
+  @type status :: String.t()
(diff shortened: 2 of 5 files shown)


core/lib/sykli/work_item.ex
```diff
@@ -0,0 +1,241 @@
...
+    if value == "", do: nil, else: value
+  end
+
+  defp blank_to_nil(value), do: value
+
+  defp now_iso8601 do
+    DateTime.utc_now()
+    |> DateTime.truncate(:second)
+    |> DateTime.to_iso8601()
+  end
```
yairfalse, line 342: **`DateTime.utc_now/0` likely violates `CredoSykli.NoWallClock`.**

CLAUDE.md is explicit:

> the custom `CredoSykli.Check.NoWallClock` check fails on `System.monotonic_time/os_time/system_time`, **`DateTime.utc_now`**, `NaiveDateTime.utc_now`, `:os.system_time`, `:erlang.now`, and bare `:rand.uniform`.

The PR description reports `mix format` + `mix test` but not `mix credo`. Could you confirm `cd core && mix credo --strict` is clean?

If it flags this function, the cleanest path is a shared `Sykli.Clock` (or `Sykli.Work.Clock`) behaviour following the same pattern the GitHub layer already uses (`core/lib/sykli/github/clock*.ex`). Then tests stay deterministic via the behaviour swap rather than threading `now:` through every call.

If you want to keep the inline impl and Credo allows scope-level opt-out, add the standard disable annotation here and document why this is a leaf wall-clock callsite.

core/lib/sykli/work_item.ex
```diff
@@ -0,0 +1,241 @@
...
+    with :ok <- validate_id(attrs["id"]),
+         :ok <- validate_title(attrs["title"]),
+         :ok <- validate_version(attrs["version"]),
+         :ok <- validate_status(attrs["status"] || "open"),
+         :ok <- validate_assignment_type(attrs["assigned_to_type"]),
+         :ok <- validate_notes(attrs["notes"] || []) do
+      {:ok,
+       %__MODULE__{
+         id: attrs["id"],
+         version: attrs["version"] || @version,
```
yairfalse: **Silent default-to-`"1"` for missing `version` is exactly what the engine just removed.**

From CLAUDE.md:

> Engine now enforces `version` strictly. … The previous silent default-to-`"1"` behavior is gone — payloads without a valid version are rejected.

Here `version: attrs["version"] || @version` quietly coerces a missing version to `"1"`, and `validate_version(nil) :ok` (line 207) lets it through. When work item v2 ships, an unversioned blob is ambiguous.

Suggest: require `version` in `from_map/1`, mirroring `Sykli.ContractSchemaVersion.fetch/1`. The `new/2` path always sets `@version` already, so existing tests don't need fixtures updated.

core/lib/sykli/work_item.ex
```diff
@@ -0,0 +1,241 @@
...
+  def validate_assignment_type(type) when type in @assignment_types, do: :ok
+  def validate_assignment_type(type), do: {:error, {:invalid_assignment_type, type}}
+
+  defp validate_title(title) when is_binary(title) do
+    if String.trim(title) == "", do: {:error, {:invalid_title, :empty}}, else: :ok
+  end
+
+  defp validate_title(title), do: {:error, {:invalid_title, title}}
+
+  defp validate_version(nil), do: :ok
```
yairfalse: Tied to the comment on line 168: `validate_version(nil)` returning `:ok` is what allows the silent coercion to `"1"`. If you tighten this to `{:error, {:missing_work_item_version, nil}}`, the `from_map/1` strictness falls out naturally.

core/lib/sykli/work_item.ex
```diff
@@ -0,0 +1,241 @@
...
+  @assignment_types ~w(member agent daemon)
+
+  @enforce_keys [:id, :title, :status, :created_at, :updated_at]
+  defstruct [
+    :id,
+    version: @version,
+    title: nil,
+    intent: nil,
+    status: "open",
+    created_by: nil,
```
yairfalse: **`created_by` shape: pick now, before sync needs a migration.**

`docs/self-hosted-coordinator.md` defines `created_by` as an actor ref `(type, id)` pair. Here it's persisted as a single colon-separated string (`"member:yair"` in the docs, `nil` in the struct). The two will diverge when Phase 6 ships and has to map one onto the other.

Two options, either is fine — but commit before persisted files exist in the wild:

- **(a)** Split locally: `created_by_type`, `created_by_id`. Mirrors the coordinator and the existing `assigned_to_type`/`assigned_to_id` pair.
- **(b)** Keep the colon-string locally, document the format formally in `docs/local-state-plane.md`, and have the daemon parse on sync.

I'd lean (a) for symmetry.

core/lib/sykli/work/store.ex
```diff
@@ -0,0 +1,159 @@
...
+  def update_status(id, status, opts \\ []) do
+    with {:ok, item} <- get(id, opts),
+         {:ok, updated} <- WorkItem.update_status(item, status, opts),
+         :ok <- save(updated, opts) do
+      {:ok, updated}
+    end
+  end
+
+  @doc "Claims a persisted work item."
+  def claim(id, assignment_type, assignment_id, opts \\ []) do
```
yairfalse, line 96: **`claim/4` is silently idempotent — and concurrent claims race.**

Two behaviors worth either fixing or pinning down with a test + moduledoc note:

1. `claim/4` doesn't check current status. An already-`claimed` work item can be re-claimed, overwriting `assigned_to_*` with no error. If "first claim wins" is the intended semantic at this layer, return `{:error, {:work_item_already_claimed, id, current_assignment}}` when `item.status != "open"`. The coordinator (Phase 6) gets DB-level atomicity anyway.

2. Atomic file write is per-call; two concurrent `claim/4` callers race at the application layer (read-read-modify-modify-write-write — last writer wins). No concurrent caller exists in Phase 1, but the next PR (CLI) introduces one.

Minimum suggestion: add a one-line moduledoc note at the top of `Sykli.Work.Store` explaining that this layer is last-write-wins and that the CLI is responsible for serializing user-initiated changes.

core/lib/sykli/work/store.ex
```diff
@@ -0,0 +1,159 @@
...
+      if path_within?(path, expanded_dir) do
+        {:ok, path}
+      else
+        {:error, {:work_item_path_escape, id}}
+      end
+    end
+  end
+
+  defp encode(%WorkItem{} = item) do
+    {:ok, Jason.encode!(WorkItem.to_map(item), pretty: true)}
```
yairfalse: Style nit — `Jason.encode/2` (non-bang) returns `{:ok, _} | {:error, _}` natively, so the `rescue` block becomes unnecessary:

```elixir
defp encode(%WorkItem{} = item) do
  case Jason.encode(WorkItem.to_map(item), pretty: true) do
    {:ok, json} -> {:ok, json}
    {:error, reason} -> {:error, {:work_item_encode_failed, reason}}
  end
end
```

Functionally identical, but avoids the bang-then-rescue pattern.

core/lib/sykli/work_item.ex
```diff
@@ -0,0 +1,241 @@
...
+
+  defp validate_assignment_id(nil), do: {:error, {:invalid_assignment_id, :empty}}
+
+  defp validate_assignment_id(id) when is_binary(id) do
+    if String.trim(id) == "", do: {:error, {:invalid_assignment_id, :empty}}, else: :ok
+  end
+
+  defp validate_assignment_id(id), do: {:error, {:invalid_assignment_id, id}}
+
+  defp validate_notes(notes) when is_list(notes), do: :ok
```
yairfalse: **`validate_notes/1` only checks list-ness.**

```elixir
defp validate_notes(notes) when is_list(notes), do: :ok
```

A persisted file containing `"notes": [1, 2, 3]` or `"notes": [{"foo": "bar"}]` will round-trip without complaint. The shape is then surfaced wholesale via `to_map/1` to whoever reads `item.notes` (CLI in PR 2, MCP later, eventually the coordinator).

Suggest tightening to per-element shape — at minimum require `id`, `body`, `created_at` keys. The shape is already known from `append_note/3`.

yairfalse (commented):
## Overview

Clean Phase 1 slice from `docs/team-mode-roadmap.md`: `Sykli.WorkItem` + `Sykli.Work.Store` writing `.sykli/work/items/<id>.json`, with 21 tests and a doc update. Scope discipline is exactly right — no CLI, no MCP, no networking — and the negative-assertion tests against `logs`/`artifacts`/`secrets` keys are a nice contract guard.

The code is well-structured. I have three semantic concerns that are worth resolving before the next slice (CLI) so persisted files don't need migrating, plus a few smaller items.

## Strengths

- **Path containment** uses `expanded_path == expanded_base or String.starts_with?(expanded_path, expanded_base <> "/")` — exactly the form CLAUDE.md mandates. The negative tests for `../escape` and `/tmp/escape` confirm the seam.
- **Atomic writes** via temp file + rename, with cleanup on failure.
- **Determinism through injection.** `now:`, `id:`, and `note_id:` are all injectable; tests use them consistently.
- **Closed enums centralized** (`statuses/0`, `assignment_types/0`) — matches the engine vocabulary modules pattern from CLAUDE.md.
- **Layered validation.** `from_map/1` validates id, title, version, status, assignment type, and notes type; `Store.save/2` re-validates id before constructing the path. Defense-in-depth.

## Top three semantic concerns

See inline comments for code-specific notes. The headline items:

### 1. `DateTime.utc_now/0` likely violates `CredoSykli.NoWallClock`

CLAUDE.md is explicit that the custom Credo check fails on `DateTime.utc_now`. The PR ran `mix format` and `mix test` but not `mix credo`. Worth confirming. See inline comment on `now_iso8601/0`.

### 2. Version policy is more permissive than the engine's

The engine just shipped strict `version` enforcement (`Sykli.ContractSchemaVersion`) — missing/empty/wrong-type/unsupported are all rejected, the silent default-to-`"1"` behavior was removed. `WorkItem.from_map/1` keeps that exact behavior locally: `version: attrs["version"] || @version` plus `validate_version(nil) = :ok`. Recommend tightening to match the engine's posture so future v2 migrations are unambiguous.

### 3. `created_by` shape diverges from the coordinator's

`docs/self-hosted-coordinator.md` defines `created_by` as an actor ref `(type, id)` pair. This PR persists it as a single colon-string (`"member:yair"`). Phase 6 sync will need to map one onto the other. Pick the convention now (I'd lean toward splitting locally to mirror the coordinator and the existing `assigned_to_type`/`assigned_to_id` pair) and document it.

## Project convention check

| Rule | Status |
|---|---|
| Path containment via `String.starts_with?(path, base <> "/")` | ✓ |
| Atomic writes | ✓ |
| Closed-enum vocabulary modules | ✓ |
| Engine version-strictness analogue | partial — see #2 |
| `mix format` | ✓ (in PR description) |
| `mix test` | ✓ (1539 tests, 0 failures) |
| `mix credo` | **not reported** — see inline comment on `now_iso8601` |
| `git diff --check` | ✓ |

## Test coverage

Good for what it covers. Low-effort additions worth considering:

- A round-trip property test (`to_map |> Jason.encode! |> Jason.decode! |> from_map == {:ok, original}`) — the project already has `:property` tests.
- A `from_map` negative for `"version": "2"` reaching it through `Store.get/2`.
- Document or change `list/1`'s halt-on-first-malformed-file behavior so the future CLI can render it cleanly.

CLI/blackbox/oracle correctly deferred to PR 2 per the description.

## Recommended actions before merge

1. Run `mix credo --strict` and report. If `now_iso8601/0` is flagged, route through a shared clock abstraction (the GitHub layer already uses `Sykli.GitHub.Clock` as a precedent).
2. Decide the version policy — strict like the engine, or document the local exception in `docs/local-state-plane.md`.
3. Pick one shape for `created_by` and document it.

Non-blocking: tighten `validate_notes/1`, replace `Jason.encode!` + `rescue` with the non-bang variant, add a moduledoc note about last-write-wins concurrency, and decide whether `claim/4` should require `status == "open"`.

Nice work — the slice is the right size and the design alignment is mostly there.

core/lib/sykli/work_item.ex
yairfalse: Fixed in `f91e5f3` using option (a): local work items now persist `created_by_type` and `created_by_id`, matching the coordinator actor-ref shape and the assignment pair.

core/lib/sykli/work_item.ex
yairfalse: Fixed in `f91e5f3`: `new/2` now writes `version: "1"`, while `from_map/1` rejects missing persisted versions with `{:missing_work_item_version, nil}` instead of silently defaulting.

core/lib/sykli/work_item.ex
yairfalse: Fixed with the same strict-version change: `validate_version(nil)` now returns `{:error, {:missing_work_item_version, nil}}`, and tests cover both direct model load and persisted store load.

core/lib/sykli/work_item.ex
yairfalse: Confirmed with `cd core && mix credo --strict`: clean, no issues. The rule is scoped, so I kept the inline leaf timestamp path and the existing `now:` injection for deterministic tests.

core/lib/sykli/work_item.ex
yairfalse: Fixed in `f91e5f3`: persisted notes now validate each element and require non-empty `id`, `body`, and `created_at`; author refs are validated when present. Added malformed-note tests.

core/lib/sykli/work/store.ex
yairfalse: Fixed in `f91e5f3`: model-level claims are first-claim-wins for non-open items, tests cover model and store re-claim rejection, and `Sykli.Work.Store` now documents the local file store concurrency limit.

core/lib/sykli/work/store.ex
yairfalse: Fixed in `f91e5f3`: switched to `Jason.encode/2` and explicit `{:error, {:work_item_encode_failed, reason}}` handling.

PR #188, opened 2026-05-08
Add local work item CLI commands

yairfalse:
## Summary

Adds the local Work Item CLI layer on top of the WorkItem/Store model from PR #187. This is local-only Team Mode: commands read and write `.sykli/work/items/<id>.json` through `Sykli.Work.Store` and do not introduce coordinator, network, daemon join, remote execution, or MCP behavior.

## Commands Added

- `sykli work create "Title" [--intent TEXT] [--actor TYPE:ID] [--json]`
- `sykli work list [--json]`
- `sykli work show <work-id> [--json]`
- `sykli work claim <work-id> [--actor TYPE:ID] [--json]`
- `sykli work note <work-id> "Body" [--actor TYPE:ID] [--json]`

All commands support `--json` using the existing CLI JSON envelope.

## JSON Output Shape

Successful object responses return `ok: true` and a local source marker, for example:

```json
{
  "ok": true,
  "data": {
    "source": "local",
    "item": {
      "id": "work_...",
      "version": "1",
      "title": "Investigate failing checkout deploy",
      "status": "open"
    }
  }
}
```

`work list --json` returns `data.items`; `work note --json` returns both `data.note` and the updated `data.item`. Errors use the existing structured error envelope.

## Oracle Coverage

Added oracle cases:

- `058` local work item JSON workflow: create, list, show, claim, note, persisted file check
- `059` invalid work item ID JSON error

## Blackbox Coverage

Added blackbox category `WORK` with:

- `WORK-001` full local workflow: create/list/show/claim/note plus persisted `.sykli/work/items/<id>.json` verification
- `WORK-002` invalid ID rejection through `sykli work show ../escape --json`

## What Changed

- Added `Sykli.CLI.Work`
- Routed `sykli work` from the main CLI
- Added public work item error wrapping/formatting
- Added focused CLI tests for local work commands and JSON output
- Added blackbox and oracle coverage
- Documented commands, local state, roadmap status, and public work error codes

## Validation

- `cd core && mix format lib/sykli/cli.ex lib/sykli/cli/work.ex lib/sykli/error.ex test/sykli/cli/work_test.exs` passed
- `python3 -m json.tool test/blackbox/dataset.json >/dev/null` passed
- `cd core && mix test test/sykli/cli/work_test.exs test/sykli/work_item_test.exs test/sykli/work/store_test.exs test/sykli/cli_test.exs test/sykli/cli/json_response_test.exs` passed: 69 tests, 0 failures
- `cd core && mix escript.build` passed
- `test/blackbox/run.sh --filter=WORK` passed: 2 passed, 0 failed, 150 skipped
- `eval/oracle/run.sh --case=058` passed
- `eval/oracle/run.sh --case=059` passed
- `cd core && mix credo --strict` passed: found no issues
- `cd core && mix test` passed: 1 property, 1556 tests, 0 failures, 1 skipped, 13 excluded
- `git diff --check` passed

Full conformance was not run because this PR does not change contract schema, SDK emission, or conformance fixtures.

## Out of Scope

- No coordinator
- No network sync
- No daemon join or heartbeat
- No run/work association yet
- No gate state
- No MCP tools
- No remote execution
- No changes to mesh behavior

## Next PR

Next layer: **Associate runs with work items**.

### DIFF
--- a/core/lib/sykli/cli/work.ex
+++ b/core/lib/sykli/cli/work.ex
@@ -13,10 +13,12 @@ defmodule Sykli.CLI.Work do
   alias Sykli.WorkItem
 
   @default_actor_type "member"
-  @default_actor_id "local"
 
   @doc """
   Runs a `sykli work ...` command and returns the intended process exit code.
+
+  `runtime_opts` are internal test hooks for store path, clocks, ids, and
+  default actor attribution. Parsed CLI flags are kept separate in `opts`.
   """
   def run(args, runtime_opts \\ []) do
     case parse(args) do
@@ -29,7 +31,7 @@ defmodule Sykli.CLI.Work do
   end
 
   defp execute({:create, title}, opts, runtime_opts) do
-    {actor_type, actor_id} = actor(opts)
+    {actor_type, actor_id} = actor(opts, runtime_opts)
 
     create_opts =
       runtime_opts
@@ -83,7 +85,7 @@ defmodule Sykli.CLI.Work do
   end
 
   defp execute({:claim, id}, opts, runtime_opts) do
-    {actor_type, actor_id} = actor(opts)
+    {actor_type, actor_id} = actor(opts, runtime_opts)
 
     claim_opts =
       runtime_opts
@@ -102,7 +104,7 @@ defmodule Sykli.CLI.Work do
   end
 
   defp execute({:note, id, body}, opts, runtime_opts) do
-    {actor_type, actor_id} = actor(opts)
+    {actor_type, actor_id} = actor(opts, runtime_opts)
 
     note_opts =
       runtime_opts
@@ -188,6 +190,14 @@ defmodule Sykli.CLI.Work do
     parse_flags(rest, [{:json, true} | opts], positionals)
   end
 
+  defp parse_flags(["--help" | rest], opts, positionals) do
+    parse_flags(rest, opts, ["--help" | positionals])
+  end
+
+  defp parse_flags(["-h" | rest], opts, positionals) do
+    parse_flags(rest, opts, ["-h" | positionals])
+  end
+
   defp parse_flags(["--intent", value | rest], opts, positionals) do
     parse_flags(rest, [{:intent, value} | opts], positionals)
   end
@@ -202,7 +212,7 @@ defmodule Sykli.CLI.Work do
         parse_flags(rest, [{:actor_type, actor_type}, {:actor_id, actor_id} | opts], positionals)
 
       {:error, reason} ->
-        {Enum.reverse(opts), Enum.reverse(positionals), reason}
+        finalize_parse_error(opts, positionals, rest, reason)
     end
   end
 
@@ -212,16 +222,27 @@ defmodule Sykli.CLI.Work do
         parse_flags(rest, [{:actor_type, actor_type}, {:actor_id, actor_id} | opts], positionals)
 
       {:error, reason} ->
-        {Enum.reverse(opts), Enum.reverse(positionals), reason}
+        finalize_parse_error(opts, positionals, rest, reason)
     end
   end
 
-  defp parse_flags([<<"--", _::binary>> = flag | _rest], opts, positionals) do
-    {Enum.reverse(opts), Enum.reverse(positionals), {:unknown_work_flag, flag}}
+  defp parse_flags([<<"--", _::binary>> = flag | rest], opts, positionals) do
+    finalize_parse_error(opts, positionals, rest, {:unknown_work_flag, flag})
   end
 
   defp parse_flags([arg | rest], opts, positionals) do
     parse_flags(rest, opts, [arg | positionals])
+  end
+
+  defp finalize_parse_error(opts, positionals, rest, reason) do
+    opts =
+      if "--json" in rest do
+        [{:json, true} | opts]
+      else
+        opts
+      end
+
+    {Enum.reverse(opts), Enum.reverse(positionals), reason}
   end
 
   defp parse_actor(value) when is_binary(value) do
@@ -243,11 +264,34 @@ defmodule Sykli.CLI.Work do
   defp validate_cli_actor_id(""), do: {:error, {:invalid_work_actor, :empty_id}}
   defp validate_cli_actor_id(_actor_id), do: :ok
 
-  defp actor(opts) do
+  defp actor(opts, runtime_opts) do
+    {default_type, default_id} = default_actor(runtime_opts)
+
     {
-      Keyword.get(opts, :actor_type, @default_actor_type),
-      Keyword.get(opts, :actor_id, @default_actor_id)
+      Keyword.get(opts, :actor_type, default_type),
+      Keyword.get(opts, :actor_id, default_id)
     }
+  end
+
+  defp default_actor(runtime_opts) do
+    {
+      Keyword.get(runtime_opts, :default_actor_type, @default_actor_type),
+      Keyword.get_lazy(runtime_opts, :default_actor_id, &default_actor_id/0)
+    }
+  end
+
+  defp default_actor_id do
+    ["SYKLI_ACTOR_ID", "USER", "USERNAME"]
+    |> Enum.find_value(fn name ->
+      case System.get_env(name) do
+        value when is_binary(value) and value != "" -> value
+        _ -> nil
+      end
+    end)
+    |> case do
+      nil -> "local"
+      value -> value
+    end
   end
 
   defp store_opts(runtime_opts), do: Keyword.take(runtime_opts, [:path])
@@ -297,6 +341,8 @@ defmodule Sykli.CLI.Work do
     Usage: sykli work <command>
 
     Local work item commands.
+
+    Unquoted title/body words are joined with spaces. Unknown flags are rejected.
 
     Commands:
       sykli work create "Title" [--intent TEXT] [--actor TYPE:ID]
--- a/core/lib/sykli/cli/work.ex
+++ b/core/lib/sykli/cli/work.ex
@@ -1,0 +1,314 @@
+defmodule Sykli.CLI.Work do
+  @moduledoc """
+  Local work item CLI commands.
+
+  This module is intentionally local-only. Team/coordinator routing is added in
+  later Team Mode phases.
+  """
+
+  alias Sykli.CLI.JsonResponse
+  alias Sykli.Error
+  alias Sykli.Error.Formatter
+  alias Sykli.Work.Store
+  alias Sykli.WorkItem
+
+  @default_actor_type "member"
+  @default_actor_id "local"
+
+  @doc """
+  Runs a `sykli work ...` command and returns the intended process exit code.
+  """
+  def run(args, runtime_opts \\ []) do
+    case parse(args) do
+      {:ok, command, opts} ->
+        execute(command, opts, runtime_opts)
+
+      {:error, reason, opts} ->
+        output_error(reason, Keyword.get(opts, :json, false))
+    end
+  end
+
+  defp execute({:create, title}, opts, runtime_opts) do
+    {actor_type, actor_id} = actor(opts)
+
+    create_opts =
+      runtime_opts
+      |> Keyword.take([:path, :now, :id])
+      |> Keyword.put(:intent, Keyword.get(opts, :intent))
+      |> Keyword.put(:created_by_type, actor_type)
+      |> Keyword.put(:created_by_id, actor_id)
+
+    case Store.create(title, create_opts) do
+      {:ok, item} ->
+        output_success(%{source: "local", item: item_map(item)}, opts, fn ->
+          IO.puts("Created work item #{item.id}")
+          IO.puts(item.title)
+        end)
+
+      {:error, reason} ->
+        output_error(reason, Keyword.get(opts, :json, false))
+    end
+  end
+
+  defp execute(:list, opts, runtime_opts) do
+    case Store.list(store_opts(runtime_opts)) do
+      {:ok, items} ->
+        output_success(%{source: "local", items: Enum.map(items, &item_map/1)}, opts, fn ->
+          if items == [] do
+            IO.puts("No work items")
+          else
+            IO.puts("Work items:")
+
+            Enum.each(items, fn item ->
+              IO.puts("  #{item.id}  #{item.status}  #{item.title}")
+            end)
+          end
+        end)
+
+      {:error, reason} ->
+        output_error(reason, Keyword.get(opts, :json, false))
+    end
+  end
+
+  defp execute({:show, id}, opts, runtime_opts) do
+    case Store.get(id, store_opts(runtime_opts)) do
+      {:ok, item} ->
+        output_success(%{source: "local", item: item_map(item)}, opts, fn ->
+          print_item(item)
+        end)
+
+      {:error, reason} ->
+        output_error(reason, Keyword.get(opts, :json, false))
+    end
+  end
+
+  defp execute({:claim, id}, opts, runtime_opts) do
+    {actor_type, actor_id} = actor(opts)
+
+    claim_opts =
+      runtime_opts
+      |> Keyword.take([:path, :now])
+
+    case Store.claim(id, actor_type, actor_id, claim_opts) do
+      {:ok, item} ->
+        output_success(%{source: "local", item: item_map(item)}, opts, fn ->
+          IO.puts("Claimed work item #{item.id}")
+          IO.puts("#{item.assigned_to_type}:#{item.assigned_to_id}")
+        end)
+
+      {:error, reason} ->
+        output_error(reason, Keyword.get(opts, :json, false))
+    end
+  end
+
+  defp execute({:note, id, body}, opts, runtime_opts) do
+    {actor_type, actor_id} = actor(opts)
+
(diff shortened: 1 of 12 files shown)


core/lib/sykli/cli/work.ex
```diff
@@ -0,0 +1,314 @@
...
+  """
+
+  alias Sykli.CLI.JsonResponse
+  alias Sykli.Error
+  alias Sykli.Error.Formatter
+  alias Sykli.Work.Store
+  alias Sykli.WorkItem
+
+  @default_actor_type "member"
+  @default_actor_id "local"
```
yairfalse: **Default actor `member:local` is about to be baked in.**

With no `--actor`, every locally-created work item gets `created_by_type: "member"`, `created_by_id: "local"`. The CLI tests assert this, so changing it later breaks the contract. Once Phase 6 sync ships, four people on four laptops will all show up in coordinator state as `member:local` — meaningless for attribution.

A few options to consider, ordered by my preference:

1. **No default — require `--actor` only when it matters for sync.** For Phase 1 (local-only) you can omit `created_by_*` entirely; on Phase 6 sync, prompt or fail with a helpful error if the user hasn't configured a default. Most friction-free for solo use, honest about provenance.
2. **`member:$USER`** (read from the OS). Tests still get a deterministic value via `runtime_opts`. Real-world users get something attributable.
3. **`system:local`** as a more honest "this came from a local CLI invocation" placeholder. Better than `member:local` because it doesn't pretend to be a member.

Whichever you pick, lock it down before this PR merges — the tests at `test/sykli/cli/work_test.exs` are now part of the contract.

core/lib/sykli/error.ex
```diff
@@ -860,6 +962,29 @@ defmodule Sykli.Error do
...
+    do: "unsupported version #{inspect(version)}"
+
+  defp format_work_item_reason({:unknown_flag, flag}), do: "unknown flag #{flag}"
+  defp format_work_item_reason({:invalid_command, ""}), do: "missing work command"
+
+  defp format_work_item_reason({:invalid_command, command}),
+    do: "invalid work command #{inspect(command)}"
+
+  defp format_work_item_reason({:invalid_actor, actor}), do: "invalid actor #{inspect(actor)}"
+  defp format_work_item_reason(reason), do: inspect(reason)
```
yairfalse, line 1019: **This catch-all renders ugly tuples for the most likely CLI errors.**

The wrap chain at lines ~756-770 squashes a lot of internal tuples into `invalid_work_item({:thing, value})`, but `format_work_item_reason/1` only handles `:missing_version`, `:unsupported_version`, `:unknown_flag`, `:invalid_command`, and `:invalid_actor`. Everything else falls through to `inspect/1`.

The practical effect: `sykli work claim work_001 --actor robot:r2d2` produces:

```
Error: invalid local work item: {:invalid_assignment_type, "robot"}
```

Not pretty for humans, and the JSON `error.message` mirrors it.

Add format clauses for at least these:

```elixir
defp format_work_item_reason({:invalid_assignment_type, type}),
  do: "invalid assignment type #{inspect(type)} (expected member, agent, or daemon)"

defp format_work_item_reason({:invalid_status, status}),
  do: "invalid status #{inspect(status)}"

defp format_work_item_reason({:invalid_assignment_id, _}),
  do: "assignment id is required"

defp format_work_item_reason({:invalid_actor_type, type}),
  do: "invalid actor type #{inspect(type)} (expected member, agent, or daemon)"

defp format_work_item_reason({:invalid_note, :empty_body}),
  do: "note body cannot be empty"

defp format_work_item_reason({:invalid_notes, _}),
  do: "notes must be a list"
```

core/lib/sykli/cli/work.ex
```diff
@@ -0,0 +1,314 @@
...
+
+      positionals in [[], ["--help"], ["-h"]] ->
+        {:ok, :help, opts}
+
+      positionals == ["create"] ->
+        {:error, :work_item_missing_title, opts}
+
+      match?(["create" | _], positionals) ->
+        ["create" | title_parts] = positionals
+        title = title_parts |> Enum.join(" ") |> String.trim()
```
yairfalse, line 146: **Implicit multi-arg join may surprise users.**

```elixir
title = title_parts |> Enum.join(" ") |> String.trim()
```

Means `sykli work create A B C` becomes `title = "A B C"`, even though all docs and help text show the title quoted. Most CLIs (`git commit -m`, `docker run --name`) require quoting and either error or take only the first positional.

Three paths:

1. **Pin the current behavior** with a unit test (`test "create joins multi-arg title"`) and document it in the help text and `docs/local-state-plane.md`. It is genuinely friendlier for unix users.
2. **Restrict to a single positional** — `["create", title | extras]` returns an error if `extras != []`. Aligns with stricter CLIs.
3. **Take the first positional only**, drop the rest. Worst of both worlds (silent ignore).

I'd lean (1) since the friendliness is real, but it does need to be a decision rather than an artifact. Same applies to `note <id> body` at line 168.

core/lib/sykli/cli/work.ex
```diff
@@ -0,0 +1,314 @@
...
+    case parse_actor(value) do
+      {:ok, actor_type, actor_id} ->
+        parse_flags(rest, [{:actor_type, actor_type}, {:actor_id, actor_id} | opts], positionals)
+
+      {:error, reason} ->
+        {Enum.reverse(opts), Enum.reverse(positionals), reason}
+    end
+  end
+
+  defp parse_flags([<<"--", _::binary>> = flag | _rest], opts, positionals) do
```
yairfalse: **Strict unknown-flag rejection diverges from the rest of the CLI.**

Project memory captures the existing convention as:

> Unknown CLI flags produce warnings, not errors (exit 0)

This branch returns `{:unknown_work_flag, flag}`, which `Error.wrap/1` turns into `invalid_work_item` with exit 1. Stricter is arguably better UX (typos fail loud), but inconsistency between subcommands is harder to reason about.

Decide which is canonical:

- **If strict is the new policy:** make it a project-wide change in a follow-up, document it in `docs/done.md` or `CLAUDE.md`, and update the existing CLI to match.
- **If permissive is the convention:** soften this to a stderr warning + continue parsing, like the rest of the CLI.

Also worth considering: when an agent passes `--unknown-flag-from-future-version`, should that fail the call or warn-and-proceed? The agent surface arguably wants stricter behavior than the human surface. A `--strict` global, or `SYKLI_STRICT_FLAGS=1`, could give both.

core/lib/sykli/cli/work.ex
```diff
@@ -0,0 +1,314 @@
...
+    {Enum.reverse(opts), Enum.reverse(positionals), {:unknown_work_flag, flag}}
+  end
+
+  defp parse_flags([arg | rest], opts, positionals) do
+    parse_flags(rest, opts, [arg | positionals])
+  end
+
+  defp parse_actor(value) when is_binary(value) do
```
yairfalse, line 248: Style nit on `parse_actor/1` plus a coverage gap.

The `with` chain returns the inner `:ok` tuple from `validate_*` and wraps the success in `{:ok, actor_type, actor_id}`. Looks correct, but if either validation returns `{:error, reason}`, that error tuple bubbles directly out of the `with` — and the caller gets `{:error, reason}` (the inner reason) rather than `{:error, {:invalid_work_actor, value}}`. Worth tracing once: the test only covers the happy path (`agent:claude`), not failure cases like `--actor xyz` (no colon), `--actor :empty` (empty type), or `--actor robot:r2d2` (unknown type).

Add at least one failing-actor test in `test/sykli/cli/work_test.exs` to pin the error code each form produces. Today they all collapse into `invalid_work_item` — see my comment on `format_work_item_reason/1` for why those messages render poorly.

yairfalse (commented):
## Overview

Clean CLI layer over PR #187's `Sykli.Work.Store`, completing the local half of Phase 1. `sykli work create / list / show / claim / note` all support `--json` with the standard envelope and `"source": "local"` marker. New error codes are cataloged in `docs/error-codes.md`. Oracle 058/059 + blackbox `WORK-001`/`WORK-002` cover happy path + path-traversal rejection.

It also resolves several concerns from the PR #187 review:

- ✅ `claim/4` is now strict-first-claim-wins (`work_item_already_claimed` error code, tested).
- ✅ `created_by` is split into `created_by_type`/`created_by_id` to mirror the coordinator data model.
- ✅ `mix credo --strict` reportedly passes — my `DateTime.utc_now/0` concern from #187 wasn't realized.

The substantive items left fall into one important semantic decision and a handful of UX/error-rendering issues. See inline comments.

## Strengths

- **Dual-surface discipline.** Every command supports `--json`, returns the shared envelope via `Sykli.CLI.JsonResponse`, and routes errors through `Sykli.Error.wrap/1`. Matches `docs/done.md` exactly.
- **`source: "local"` marker** in every JSON payload. Sets up Phase 6 sync to differentiate `local` vs `coordinator` cleanly.
- **stdout vs stderr split done right.** Human errors → stderr, JSON envelopes (success or error) → stdout. Agents pipe stdout to jq cleanly.
- **Global `--json` normalization** extended for the new subcommand (`Sykli.CLI.normalize_global_json(["--json", "work", "list"]) == ["work", "--json", "list"]`), so both `sykli --json work list` and `sykli work list --json` produce identical output.
- **Error catalog updated** in `docs/error-codes.md` under a new `work` section, all marked `public-unstable` (correct tier for Phase 1).
- **Oracle + blackbox coverage** for the binary-level workflow, including a path-traversal negative.

## Top concerns

See inline comments for code-anchored notes. The headline items:

### 1. Default actor `member:local` will pollute coordinator state at Phase 6

`@default_actor_type "member"` + `@default_actor_id "local"` means every locally-created work item gets `created_by_type: "member", created_by_id: "local"` unless the user remembers `--actor`. Tests bake this in. When Phase 6 syncs, four people on four laptops will all show up as `member:local` — useless for attribution. Consider `system:local`, `member:$USER`, or — better — no default until Phase 6 forces the question.

### 2. `format_work_item_reason/1` falls through to `inspect/1` for the most likely CLI errors

The wrap chain squashes specific tuples into the umbrella `invalid_work_item` error. `format_work_item_reason/1` only handles a few of them; the rest render as raw Elixir tuples. `sykli work claim work_001 --actor robot:r2d2` produces `"invalid local work item: {:invalid_assignment_type, \"robot\"}"`. Add format clauses for `:invalid_assignment_type`, `:invalid_status`, `:invalid_assignment_id`, `:invalid_actor_type`, `:invalid_note`, `:invalid_notes`.

### 3. Title/body multi-arg join is implicit

`sykli work create A B C` joins to `"A B C"`. Users without quotes will Just Work — but it's surprising vs `git commit "A" "B"` (most CLIs error or take only the first arg). Either pin the behavior with a test + doc, or restrict to a single positional. Same for `note`.

### 4. Unknown-flag handling diverges from the rest of the CLI

The project convention is "unknown CLI flags produce warnings, not errors (exit 0)". The `work` CLI rejects unknown flags with exit 1. Strict rejection is arguably better UX; either way, this divergence should be intentional and documented.

## Project convention check

| Rule | Status |
|---|---|
| `source: "local"` per design | ✓ |
| Error codes added to `docs/error-codes.md` | ✓ (line numbers in `Emitted from` column missing — see below) |
| `mix format` / `mix credo --strict` | ✓ (claimed in PR description) |
| `mix test` | ✓ (1556 tests, 0 failures) |
| `git diff --check` | ✓ |
| Unknown-flag convention | ⚠ — see issue #4 |
| MCP counterpart for new CLI | not in this PR; explicitly deferred. Phase 1 exit criteria (`docs/team-mode-roadmap.md`) calls for MCP tools — confirm there's a follow-up issue tracking that. |

## Test coverage

Good for happy paths and the core error codes. Gaps worth filling either here or in the next PR:

- `--intent=value` and `--actor=value` (equals-sign forms parse successfully — only the space form is tested).
- Unknown-subcommand: `sykli work foo` (today: `:invalid_work_command` → ugly inspect render — see issue #2).
- Unknown-flag: `sykli work list --bogus` (today: exit 1 — see issue #4).
- `--help` exit 0 path.
- A round-trip property test on the `--json` envelope shape.

The blackbox `WORK-001` is one chained `&&` workflow — works, but a failure in step 3 reports as "command failed" without telling you which step. Consider splitting into per-command cases for diagnostic clarity, or add a wrapper script that echoes the failing step.

Oracle 058 and blackbox `WORK-001` cover the same workflow. That's fine — they serve different evaluation purposes — but worth noting the redundancy.

## Security

- Path containment is preserved (delegated to `Store` from PR #187). The blackbox `WORK-002` and oracle 059 confirm `../escape` is rejected. ✓
- `--actor TYPE:ID` validates `TYPE` against the closed enum and rejects empty `ID`, but `ID` is otherwise free-form text. It flows into stored JSON `created_by_id`/`assigned_to_id`/`author_id` and is never used as a path component. Safe.
- Note `body` is plaintext, no escaping at write time. Future MCP/coordinator surfaces need to consider rendering implications, but that's out of scope here.
- No new attack surface beyond PR #187. ✓

## Minor nits

- `docs/error-codes.md` entries lack line numbers in the `Emitted from` column — existing entries in other categories include them (e.g. `core/lib/sykli/error.ex:468`).
- `README.md` places `work/items/` before `runs/` in the `.sykli/` tree; existing entries seem grouped by purpose, so this may be fine, just noting.
- The `runtime_opts` vs `opts` parameter naming in `Sykli.CLI.Work` is internally consistent but slightly opaque on first read. A one-line comment on `run/2` documenting that `runtime_opts` is for test injection (path/now/id/note_id) and `opts` is for user-supplied flags would help.
- `Store.claim/4`'s last-write-wins concurrency limitation from PR #187 is still present. The CLI doesn't add file-level serialization. Phase 6 sync gets DB-level atomicity; for the local CLI today, two concurrent `sykli work claim <id>` racers can still both succeed. Not blocking; worth a moduledoc note.

## Bottom line

Well-scoped Phase 1 CLI slice with strong dual-surface discipline. The default-actor decision is the only thing I'd want resolved before more PRs land on top — once `member:local` is in everyone's `.sykli/work/items/`, future Phase 6 sync inherits the noise. The error-rendering fall-through and unknown-flag divergence are smaller follow-ups.

yairfalse (commented):
## Follow-up review of `c9c5523` — "address work command review feedback"

All four substantive concerns from the prior review are resolved. Approving.

### Resolved

**1. Default actor (was: `member:local` would pollute coordinator state).** Replaced with an env-var fallback chain: `SYKLI_ACTOR_ID` → `USER` → `USERNAME` → `"local"`. Tests inject `default_actor_id: "test-user"` via the new `runtime_opts` seam, so determinism is preserved and production picks up real attribution. `USERNAME` covers Windows; explicit override beats both. Good design.

**2. Error rendering (was: `format_work_item_reason/1` falls through to `inspect/1`).** Eleven new format clauses cover `:invalid_status`, `:invalid_assignment_type`, `:invalid_assignment_id`, `:invalid_actor_type`, `:invalid_actor_id`, `:invalid_created_by`, `:invalid_notes`, `:invalid_note`, `:invalid_note_author`, `:empty_id`, `:empty_body`, `:not_string`, `:missing_type`. The new test

```elixir
refute result["error"]["message"] =~ "{:"
```

is a great regression guard — any future internal tuple that leaks through the umbrella error without a format clause will fail this assertion.

**3. Multi-arg title/body join (was: implicit and surprising).** Now explicitly pinned: help text declares `"Unquoted title/body words are joined with spaces."`, and two new tests assert the joining behavior for both `create` and `note`. The behavior is now part of the contract.

**4. Unknown-flag rejection (was: divergence from CLI-wide convention).** Help text declares `"Unknown flags are rejected."`, and a test pins the exact error message. The new `finalize_parse_error/4` also propagates `--json` across error paths so `sykli work list --bogus --json` correctly returns a JSON envelope instead of human text — small but important fix for the agent surface.

### Resolved minor nits

- `docs/error-codes.md` entries now carry line numbers (`error.ex:622` etc.).
- `runtime_opts` vs `opts` is documented in the moduledoc.
- `--help` exit-0 path covered (`"help exits successfully"`).
- `--intent=value` and `--actor=value` equals-forms covered.
- Unknown subcommand (`sykli work wat`) covered.

### Remaining minor items (not blocking)

- **`default_actor_id/0` env-var chain has no direct unit test** — only the `runtime_opts` injection seam is tested. A small synchronized test using `System.put_env`/`delete_env` would prove the priority order (`SYKLI_ACTOR_ID > USER > USERNAME > "local"`). Optional; the seam is solid.
- **`sykli work create --help "title"` produces a work item titled `"--help title"`** because `--help` is now treated as a positional and the multi-arg join concatenates it. The user can always run `sykli work --help` to recover, so this is more curiosity than bug.
- **`Store.claim/4` last-write-wins concurrency** carry-over from PR #187 still uncovered. Phase 6 sync gives DB-level atomicity; a moduledoc note on `Sykli.Work.Store` would still be worth adding before more PRs stack.
- README ordering of `work/items/` before `runs/` in the `.sykli/` tree — trivial cosmetic.

### Bottom line

Thorough, focused response to feedback. The env-var actor chain, the regression-prevention assertion against tuple leakage, and the `finalize_parse_error/4` JSON-context fix are all genuinely good additions on top of just "answering the comments." Phase 1's local CLI is in good shape.

LGTM from my end — leaving as a comment review since GitHub blocks approving one's own PR.

PR #189, opened 2026-05-08
Associate runs with work items

yairfalse:
## Summary

This PR connects local Team Mode work items to actual Sykli runs. A local run can now be associated with a work item using `sykli run --work <work-id>`, and associated runs can be listed with `sykli work runs <work-id>`.

The implementation stays local-first: it stores the relationship in existing `.sykli/runs/*.json` run manifests and does not introduce coordinator, network sync, daemon join, MCP, or remote execution behavior.

## Why

A Work Item is the team-level coordination unit. A Run is one execution of one contract. A work item needs durable run references before coordinator sync can exist, and the local `.sykli/` state needs to remain the detailed source of truth.

## What changed

- Added deterministic `Sykli.ContractHash` helpers using `sha256:` over the detected SDK file bytes.
- Added optional `work_item_id` and `contract_hash` fields to run history manifests.
- Added `Sykli.RunHistory.list_by_work_item/2` for deterministic recent-first listing.
- Added `sykli run --work <work-id>` validation before execution starts.
- Added `work_item_id`, `contract_hash`, and `run_id` to run JSON output when `--work` is used.
- Added `sykli work runs <work-id>` with human and `--json` output.
- Added `contract_hash_failed` to the public error catalog.
- Updated README and Team Mode docs for the local work/run flow.

## Validation

- `python3 -m json.tool test/blackbox/dataset.json >/dev/null` — passed
- `git diff --check` — passed
- `cd core && mix format --check-formatted` — passed
- `cd core && mix test test/sykli/contract_hash_test.exs test/sykli/run_history_test.exs test/sykli/cli/work_test.exs test/sykli/cli_test.exs test/sykli/work/run_association_test.exs` — passed, 71 tests, 0 failures
- `cd core && mix escript.build` — passed
- `eval/oracle/run.sh --case=060` — passed
- `eval/oracle/run.sh --case=061` — passed
- `eval/oracle/run.sh --case=062` — passed
- `test/blackbox/run.sh --filter=WORK` — passed, 5 passed, 0 failed, 150 skipped
- `cd core && mix test` — passed, 1 property, 1574 tests, 0 failures, 1 skipped, 13 excluded

Full conformance was not run because this PR does not change contract schemas, SDK emitters, or conformance JSON output.

## Oracle coverage

Added oracle cases:

- `060` — `sykli run --work <id> --json` includes `work_item_id`, `contract_hash`, and `status`.
- `061` — `sykli work runs <id> --json` lists the associated run with a stable `sha256:` hash.
- `062` — missing work item returns structured JSON error `work_item_not_found` before execution.

## Blackbox coverage

Added blackbox cases:

- `WORK-003` — create work item, run with `--work`, list associated runs, and verify work item state does not copy task output.
- `WORK-004` — missing work item fails before execution.
- `WORK-005` — invalid/traversal-like work item id fails before execution.

## Out of scope

- No coordinator
- No networking
- No daemon join or heartbeat
- No local gate decision state
- No MCP tools
- No Kubernetes deployment
- No remote execution
- No SaaS or web UI

## Next PR

Add local gate decision state:

- local gate model/store
- `sykli gates list`
- `sykli gate show`
- `sykli gate approve`
- `sykli gate reject`
- `--json` output
- gate oracle coverage
- gate blackbox workflow

### DIFF
--- a/core/lib/sykli/cli.ex
+++ b/core/lib/sykli/cli.ex
@@ -7,6 +7,7 @@ defmodule Sykli.CLI do
   alias Sykli.Delta
   alias Sykli.Error
   alias Sykli.Error.Formatter
+  alias Sykli.Work.Store, as: WorkStore
 
   @version Mix.Project.config()[:version]
   @subcommands ~w(cache graph delta watch report history daemon work init validate verify plan explain context fix query mcp run)
@@ -111,6 +112,7 @@ defmodule Sykli.CLI do
       --filter=NAME             Only run tasks matching NAME
       --timeout=DURATION        Per-task timeout (default: 5m). Use 10m, 30s, 2h, 1d, or 0 for no timeout
       --target=TARGET           Execution target: local (default), k8s
+      --work=WORK_ID            Associate this run with a local work item
       --allow-dirty             Allow running with uncommitted changes (K8s)
       --git-ssh-secret=NAME     K8s Secret for SSH git auth
       --git-token-secret=NAME   K8s Secret for HTTPS git auth
@@ -237,16 +239,35 @@ defmodule Sykli.CLI do
         run_opts
       end
 
-    result = Sykli.run(path, run_opts)
-    duration = System.monotonic_time(:millisecond) - start_time
+    case prepare_work_run_context(path, opts) do
+      {:ok, work_meta, work_run_opts} ->
+        run_opts = work_run_opts ++ run_opts
+        result = Sykli.run(path, run_opts)
+        duration = System.monotonic_time(:millisecond) - start_time
 
-    # Restore stdout before writing output
-    if original_gl, do: restore_stdout(original_gl)
+        # Restore stdout before writing output
+        if original_gl, do: restore_stdout(original_gl)
 
+        output_run_result(result, duration, path, opts, json_output, work_meta)
+
+      {:error, reason} ->
+        if original_gl, do: restore_stdout(original_gl)
+
+        if json_output do
+          IO.puts(JsonResponse.error(Error.wrap(reason)))
+        else
+          display_error(reason)
+        end
+
+        halt(1)
+    end
+  end
+
+  defp output_run_result(result, duration, path, opts, json_output, work_meta) do
     case result do
       {:ok, results} ->
         if json_output do
-          IO.puts(JsonResponse.ok(run_json_data("passed", results, duration, path)))
+          IO.puts(JsonResponse.ok(run_json_data("passed", results, duration, path, work_meta)))
         else
           IO.puts(
             Renderer.render_run(
@@ -264,7 +285,7 @@ defmodule Sykli.CLI do
 
       {:error, results} when is_list(results) ->
         if json_output do
-          IO.puts(JsonResponse.ok(run_json_data("failed", results, duration, path)))
+          IO.puts(JsonResponse.ok(run_json_data("failed", results, duration, path, work_meta)))
         else
           IO.puts(
             Renderer.render_run(
@@ -291,13 +312,42 @@ defmodule Sykli.CLI do
     end
   end
 
-  defp run_json_data(status, results, duration_ms, path) do
+  defp prepare_work_run_context(path, opts) do
+    case Keyword.get(opts, :work_item_id) do
+      nil ->
+        {:ok, %{}, []}
+
+      work_item_id ->
+        with {:ok, _item} <- WorkStore.get(work_item_id, path: path),
+             {:ok, {sdk_file, _runner}} <- Sykli.Detector.find(path),
+             {:ok, contract_hash} <- Sykli.ContractHash.from_sdk_file(sdk_file) do
+          work_meta = %{work_item_id: work_item_id, contract_hash: contract_hash}
+          {:ok, work_meta, [work_item_id: work_item_id, contract_hash: contract_hash]}
+        end
+    end
+  end
+
+  defp run_json_data(status, results, duration_ms, path, work_meta) do
     %{
       status: status,
       duration_ms: duration_ms,
       tasks: Enum.map(results, &task_result_to_map/1),
       occurrence_path: Path.join([path, ".sykli", "occurrence.json"])
     }
+    |> Map.merge(latest_run_json_meta(path, work_meta))
+  end
+
+  defp latest_run_json_meta(_path, work_meta) when map_size(work_meta) == 0, do: %{}
+
+  defp latest_run_json_meta(path, work_meta) do
+    case Sykli.RunHistory.load_latest(path: path) do
+      {:ok, run} ->
+        work_meta
+        |> Map.put(:run_id, run.id)
+
+      {:error, _} ->
+        work_meta
+    end
   end
 
   defp pipeline_label(path) do
@@ -446,6 +496,13 @@ defmodule Sykli.CLI do
   defp do_parse_run_args(["--target", value | tail], opts, rest)
        when is_binary(value) and binary_part(value, 0, 1) != "-",
        do: do_parse_run_args(tail, [{:target, parse_target(value)} | opts], rest)
+
+  defp do_parse_run_args(["--work=" <> work_item_id | tail], opts, rest),
+    do: do_parse_run_args(tail, [{:work_item_id, work_item_id} | opts], rest)
+
+  defp do_parse_run_args(["--work", work_item_id | tail], opts, rest)
+       when is_binary(work_item_id) and binary_part(work_item_id, 0, 1) != "-",
+       do: do_parse_run_args(tail, [{:work_item_id, work_item_id} | opts], rest)
 
   defp do_parse_run_args(["--" <> _ = arg | tail], opts, rest) do
     IO.puts(
--- a/core/lib/sykli/cli.ex
+++ b/core/lib/sykli/cli.ex
@@ -319,10 +319,13 @@ defmodule Sykli.CLI do
 
       work_item_id ->
         with {:ok, _item} <- WorkStore.get(work_item_id, path: path),
-             {:ok, {sdk_file, _runner}} <- Sykli.Detector.find(path),
-             {:ok, contract_hash} <- Sykli.ContractHash.from_sdk_file(sdk_file) do
+             {:ok, sdk_file} <- Sykli.Detector.find(path),
+             {:ok, json} <- Sykli.Detector.emit(sdk_file),
+             {:ok, contract_hash} <- Sykli.ContractHash.from_json(json) do
           work_meta = %{work_item_id: work_item_id, contract_hash: contract_hash}
-          {:ok, work_meta, [work_item_id: work_item_id, contract_hash: contract_hash]}
+
+          {:ok, work_meta,
+           [work_item_id: work_item_id, contract_hash: contract_hash, emitted_json: json]}
         end
     end
   end
@@ -340,13 +343,16 @@ defmodule Sykli.CLI do
   defp latest_run_json_meta(_path, work_meta) when map_size(work_meta) == 0, do: %{}
 
   defp latest_run_json_meta(path, work_meta) do
+    # Local CLI compatibility: thread run_id through Sykli.run/2 when the
+    # executor result shape is normalized instead of reading latest.json back.
     case Sykli.RunHistory.load_latest(path: path) do
       {:ok, run} ->
         work_meta
+        |> Map.put(:source, "local")
         |> Map.put(:run_id, run.id)
 
       {:error, _} ->
-        work_meta
+        Map.put(work_meta, :source, "local")
     end
   end
 
@@ -503,6 +509,9 @@ defmodule Sykli.CLI do
   defp do_parse_run_args(["--work", work_item_id | tail], opts, rest)
        when is_binary(work_item_id) and binary_part(work_item_id, 0, 1) != "-",
        do: do_parse_run_args(tail, [{:work_item_id, work_item_id} | opts], rest)
+
+  defp do_parse_run_args(["--work" | tail], opts, rest),
+    do: do_parse_run_args(tail, [{:work_item_id, ""} | opts], rest)
 
   defp do_parse_run_args(["--" <> _ = arg | tail], opts, rest) do
     IO.puts(
--- a/core/lib/sykli/cli/work.ex
+++ b/core/lib/sykli/cli/work.ex
@@ -123,6 +123,33 @@ defmodule Sykli.CLI.Work do
     end
   end
 
+  defp execute({:runs, id}, opts, runtime_opts) do
+    store_opts = store_opts(runtime_opts)
+    history_opts = Keyword.merge(store_opts, Keyword.take(runtime_opts, [:limit]))
+
+    with {:ok, _item} <- Store.get(id, store_opts),
+         {:ok, runs} <- Sykli.RunHistory.list_by_work_item(id, history_opts) do
+      output_success(
+        %{source: "local", work_item_id: id, runs: Enum.map(runs, &run_map/1)},
+        opts,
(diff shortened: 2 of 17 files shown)


core/lib/sykli/contract_hash.ex
```diff
@@ -0,0 +1,26 @@
...
+  @moduledoc """
+  Deterministic hashes for emitted Sykli contracts.
+
+  Phase 1 Team Mode uses the SDK source file bytes as the local contract
+  identity. This keeps the hash stable and avoids including runtime data such
+  as timestamps, durations, or run ids.
+  """
+
+  @doc "Computes a sha256-prefixed hash for a detected SDK file."
+  def from_sdk_file(path) when is_binary(path) do
```
yairfalse: **Hash input choice diverges from the documented coordinator data model.**

`docs/self-hosted-coordinator.md` (the `contracts` table) says:

> | `contract_hash` | text | sha256 of canonical JSON |

This module hashes the SDK source file bytes instead. The moduledoc above acknowledges this is a Phase-1 simplification, but it has real downstream implications:

- Two daemons running the same emitted contract via SDK files with different formatting/comments produce **different** hashes.
- A no-op edit to `sykli.exs` (whitespace, trailing newline, comment) produces a new hash even though the contract didn't change.
- An SDK with non-deterministic source rendering (rare) would produce hashes that drift even on the same machine.

The values land in `.sykli/runs/*.json` today and are cataloged in `docs/error-codes.md` as `public-unstable`, so changing them later is allowed — but every persisted hash from this PR is one that Phase 6 sync may eventually upload.

Two paths to align:

- **(a)** Update `docs/self-hosted-coordinator.md` to describe the local hash as "sha256 of the detected SDK file bytes" and add a Phase 6 reconciliation note (e.g. "the daemon recomputes the canonical-JSON hash at sync time").
- **(b)** Hash the emitted canonical JSON instead. `Sykli.run/2` (or the SDK runner inside it) already produces this; threading it back to `prepare_work_run_context` is a small change and matches the design doc.

I'd lean (b). Either way, please pin the choice in the doc and the moduledoc so future PRs know which is canonical.

core/lib/sykli/run_history.ex
```diff
@@ -223,6 +227,17 @@ defmodule Sykli.RunHistory do
...
   end
 
+  @doc """
+  Lists recent runs associated with a local work item.
+  """
+  @spec list_by_work_item(String.t(), keyword()) :: {:ok, [Run.t()]} | {:error, term()}
+  def list_by_work_item(work_item_id, opts \\ []) do
+    with :ok <- Sykli.WorkItem.validate_id(work_item_id),
+         {:ok, runs} <- list(opts) do
+      {:ok, Enum.filter(runs, &(&1.work_item_id == work_item_id))}
```
yairfalse: **Filter happens after the limit, not before.**

```elixir
with :ok <- Sykli.WorkItem.validate_id(work_item_id),
     {:ok, runs} <- list(opts) do
  {:ok, Enum.filter(runs, &(&1.work_item_id == work_item_id))}
end
```

`list/1` already accepts a `:limit`. If the caller passes `limit: 10` and only 4 of the most recent 10 runs belong to this work item, the result is 4 runs — even when there are older runs for this work item that the user probably wanted to see.

For Phase 2 quantities this rarely matters, but `sykli work runs <id> --json` will hit it once a few unrelated runs accumulate.

Two options:

- Drop `:limit` for this code path (load all, filter, then truncate). O(N) reads but correct.
- Stream lazily: enumerate manifest filenames newest-first, decode, filter, take up to `limit`.

The lazy approach is more work but scales. For now, dropping `:limit` and adding a moduledoc note ("returns all runs associated with the work item; pagination is not yet implemented") is the lower-risk fix.

yairfalse (commented):
## Overview

Clean Phase 2 slice: `sykli run --work <id>` validates the work item before execution and stamps `work_item_id` + `contract_hash` onto the run history manifest; `sykli work runs <id>` lists associated runs. New `Sykli.ContractHash` module, new public error `contract_hash_failed`, new `RunHistory.list_by_work_item/2` helper. Three oracle cases (060–062) and three blackbox cases (`WORK-003/004/005`) cover the happy path and pre-execution failure modes.

Two strong things stand out:

- **Validation happens before execution.** `prepare_work_run_context` in `cli.ex` resolves the work item and contract hash *before* `Sykli.run/2` starts, so `--work missing` produces a clean error instead of a half-completed run. WORK-004/005 prove this.
- **`WORK-003` asserts `! grep -q "hello"` against the work item file.** Quietly enforces the design rule that work item files store coordination state, never run output. Excellent regression guard for the local-state-plane boundary.

One **important design alignment** worth resolving before persisted hashes proliferate, plus a few smaller items. See inline comments.

## Top concern: `contract_hash` input diverges from the design doc

`docs/self-hosted-coordinator.md` defines:

| `contract_hash`   | text       | sha256 of canonical JSON                    |

This PR hashes **SDK source file bytes** instead. `Sykli.ContractHash` moduledoc acknowledges this is a Phase-1 simplification, but the implication is real: when Phase 6 sync ships, two daemons running the same emitted JSON via SDK files that differ in formatting/comments will appear to the coordinator as different contracts. That's a future bug, not a current one — but the persisted hashes from this PR are about to populate `.sykli/runs/*.json` files that may later sync up.

Two paths:

- **(a)** Update the design doc to allow source-file hashing as the local-only definition, and document the Phase 6 reconciliation strategy (e.g. "the daemon recomputes a canonical hash at sync time").
- **(b)** Hash the emitted canonical JSON instead — the SDK already produces this during `sykli run`, so the bytes are available.

I'd lean (b) since it preserves the design contract, but (a) is acceptable as long as the docs and the implementation agree. Pick now to avoid migration when older `.sykli/runs/*.json` files start showing up at the coordinator.

## Strengths

- **Pre-execution validation chain** in `prepare_work_run_context/2` is concise and bubbles distinct error codes (`work_item_not_found`, `invalid_work_item_id`, `contract_hash_failed`).
- **Backward compat for runs without `--work`.** Encoder uses `maybe_add` so existing run manifests remain unchanged shape-wise; only runs with `--work` carry the new keys. The test `"normal runs without work metadata preserve previous history shape"` directly verifies this.
- **`list_by_work_item/2` validates the id first** via `Sykli.WorkItem.validate_id/1` — same path-traversal protection the work store uses.
- **Roadmap honesty.** `docs/team-mode-roadmap.md` Phase 2 update explicitly notes that `sykli explain`/`sykli fix` integration is a follow-up, not delivered here. Better than overclaiming the phase is done.
- **Error catalog updated** with `contract_hash_failed` (public-unstable) and a line number.

## Project convention check

| Rule | Status |
|---|---|
| Path containment | ✓ (relies on existing `validate_id`/`Store.get`) |
| Error codes added to `docs/error-codes.md` | ✓ with line number |
| `mix format --check-formatted` | ✓ (in PR description) |
| `mix test` | ✓ (1574 tests, 0 failures) |
| `mix credo --strict` | **not in validation list** — worth confirming |
| `git diff --check` | ✓ |
| `source: "local"` marker on run JSON output | ✗ (only on `work runs` output) — see inline comment |

## Test coverage

Good for the happy paths and most negatives. Gaps worth filling here or in a follow-up:

- **`--work --json`** (flag without a value, then a flag-looking next arg). The `do_parse_run_args` guard `binary_part(work_item_id, 0, 1) != "-"` rejects this, but it's not asserted by a test.
- **`contract_hash_failed` reaching the CLI.** The hash failure path is covered at the module level (`from_sdk_file` with a non-existent file), but no end-to-end CLI test verifies the error code propagates through `Error.wrap/1`.
- **`Detector.find` failing while `--work` is set** (e.g. `--work` in a directory with no SDK file). The `prepare_work_run_context` `with` chain leaks the detector's error through `Error.wrap/1`; verify the user sees the same error code they'd see without `--work`.
- **`list_by_work_item/2` with `:limit`** — the filter happens *after* the limit, which surfaces fewer runs than expected when older runs match. See inline comment.
- **`run_map/1` `timestamp` field** isn't asserted in the work CLI test, only `id` and `contract_hash`.

## Security

- `validate_id/1` invoked on every `--work` value before any path or store operation. WORK-005 confirms `../escape` is rejected. ✓
- `ContractHash.from_sdk_file` reads `sdk_file` returned by `Detector.find`, which already vets the path within the project. No new attack surface.
- The `contract_hash_failed` error's `message` includes the path. Minor info-leak in error text, but it's a local CLI error — practical impact low.
- Run history JSON gains two optional fields, both nullable, both opaque strings. No injection risk.

No new security issues.

## Performance

- One additional `File.read/1` + `:crypto.hash(:sha256, ...)` per `--work` invocation. Negligible.
- `list_by_work_item/2` is O(N) over all run files. Fine for typical local quantities; would want indexing if it ever spans thousands of runs, but that's beyond Phase 2.

## Bottom line

Well-scoped Phase 2 implementation with strong pre-execution validation and a quietly excellent contract-boundary assertion in WORK-003. The contract-hash input choice is the only thing I'd nail down before persisted hashes accumulate; the smaller items can ride in a follow-up. LGTM once the hash-input question is resolved.

core/lib/sykli/cli.ex
```diff
@@ -291,13 +312,42 @@ defmodule Sykli.CLI do
...
       tasks: Enum.map(results, &task_result_to_map/1),
       occurrence_path: Path.join([path, ".sykli", "occurrence.json"])
     }
+    |> Map.merge(latest_run_json_meta(path, work_meta))
+  end
+
+  defp latest_run_json_meta(_path, work_meta) when map_size(work_meta) == 0, do: %{}
+
+  defp latest_run_json_meta(path, work_meta) do
+    case Sykli.RunHistory.load_latest(path: path) do
```
yairfalse, line 348: **Race-prone read-back of `run_id` after `Sykli.run`.**

```elixir
case Sykli.RunHistory.load_latest(path: path) do
  {:ok, run} ->
    work_meta
    |> Map.put(:run_id, run.id)
```

`Sykli.run/2` writes the run manifest, then this code re-reads `RunHistory.load_latest/1` to extract the just-written `run_id`. Works in practice for single-process invocations, but:

- If two `sykli run` invocations target the same project dir concurrently, this could surface the wrong run id.
- The CLI now has implicit ordering coupling: any future change that defers manifest writes (e.g., write at SIGTERM drain time) would break the JSON output silently.

Cleaner: have `Sykli.run/2` return the `run_id` as part of its result (`{:ok, run_id, results}` or `{:ok, %{id: ..., results: [...]}}`) and pass it through to `output_run_result/6`. That's a wider refactor than this PR should take on, but a `# TODO: thread run_id through Sykli.run instead of reading it back` here would mark the seam for later.

Not blocking for Phase 2.

core/lib/sykli/cli.ex
```diff
@@ -291,13 +312,42 @@ defmodule Sykli.CLI do
...
+      nil ->
+        {:ok, %{}, []}
+
+      work_item_id ->
+        with {:ok, _item} <- WorkStore.get(work_item_id, path: path),
+             {:ok, {sdk_file, _runner}} <- Sykli.Detector.find(path),
+             {:ok, contract_hash} <- Sykli.ContractHash.from_sdk_file(sdk_file) do
+          work_meta = %{work_item_id: work_item_id, contract_hash: contract_hash}
+          {:ok, work_meta, [work_item_id: work_item_id, contract_hash: contract_hash]}
+        end
```
yairfalse, line 329: **Consider adding `source: "local"` to the run JSON output when `--work` is used.**

The Phase 1 work CLI commands (`sykli work create/list/show/claim/note`) all return `{ok: true, data: {source: "local", ...}}`. `sykli work runs <id>` continues that pattern in this PR.

But `sykli run --work <id> --json` returns the existing run JSON shape — `{status, duration_ms, tasks, occurrence_path, work_item_id, contract_hash, run_id}` — without a `source` marker.

When Phase 6 sync ships and `sykli run --team` (or similar) starts surfacing coordinator-side run state, you'll want a way to distinguish local vs coordinator runs in the JSON. Adding `source: "local"` here keeps the agent surface consistent and reserves the field for the future split.

Not blocking; a small tweak to `run_json_data/5` would land it.

core/lib/sykli/cli/work.ex
```diff
@@ -350,11 +384,25 @@ defmodule Sykli.CLI.Work do
       sykli work show <work-id>
       sykli work claim <work-id> [--actor TYPE:ID]
       sykli work note <work-id> "Body" [--actor TYPE:ID]
+      sykli work runs <work-id>
 
     Options:
       --json          Output as JSON
       --intent TEXT   Set work item intent when creating
       --actor TYPE:ID Set actor identity, e.g. member:yair, agent:claude, daemon:worker-1
```
yairfalse, line 392: Coverage gap nit: `run_map/1` produces a `timestamp` ISO8601 string, but the `runs --json` test only asserts `id`, `contract_hash`, and ordering — not the `timestamp` field shape.

A single line in the existing test would pin the contract:

```elixir
assert hd(result["data"]["runs"])["timestamp"] == "2026-05-08T11:00:00Z"
```

Low priority — the field is straightforward — but it'd catch a future change that accidentally drops or reformats the field.

core/lib/sykli/contract_hash.ex
yairfalse: Fixed in ffe2196. I switched contract_hash to use canonicalized emitted JSON rather than SDK source bytes, and Sykli.run now reuses that emitted JSON so the hash matches the contract actually executed instead of re-emitting independently. Updated the ContractHash moduledoc and local-state/roadmap docs to state canonicalized emitted JSON.

core/lib/sykli/run_history.ex
yairfalse: Fixed in ffe2196. list_by_work_item/2 now loads all local run manifests, filters by work item first, then applies :limit, with a test covering unrelated recent runs hiding older matching runs. The docstring also notes that pagination is not implemented yet.

core/lib/sykli/cli.ex
yairfalse: Addressed in ffe2196 with an explicit seam comment. I kept the result-shape refactor out of this PR, but marked that run_id should be threaded through Sykli.run/2 when executor/run result normalization happens instead of reading latest.json back.

core/lib/sykli/cli.ex
yairfalse: Fixed in ffe2196. Associated run JSON now includes source: "local", and oracle case 060 plus WORK-003 assert it.

core/lib/sykli/cli/work.ex
yairfalse: Fixed in ffe2196. The work CLI test now asserts the ISO8601 timestamp string for the newest associated run.

yairfalse:
Review feedback addressed in ffe2196.

Changes:
- `contract_hash` now hashes canonicalized emitted JSON, not SDK source bytes.
- `Sykli.run/2` reuses the emitted JSON from the CLI preflight so hash and execution use the same emitted contract.
- `list_by_work_item/2` now filters before applying `:limit`.
- `sykli run --work ... --json` now includes `source: "local"`.
- Added timestamp assertion for `work runs` JSON.
- Added parser coverage for missing `--work` value.
- Updated docs/oracle/blackbox expectations.

Validation run locally after the fix:
- `python3 -m json.tool test/blackbox/dataset.json >/dev/null` passed
- `git diff --check` passed
- `cd core && mix format --check-formatted` passed
- focused tests passed: 73 tests, 0 failures
- `cd core && mix escript.build` passed
- oracle 060/061/062 passed
- `test/blackbox/run.sh --filter=WORK` passed: 5 passed, 0 failed
- `cd core && mix credo --strict` passed: no issues
- `cd core && mix test` passed: 1 property, 1576 tests, 0 failures, 1 skipped

PR #190, opened 2026-05-09
Add local gate decision state

yairfalse:
## Summary

This PR adds the local Team Mode gate decision layer. Gates now have a versioned local model, a file-backed `.sykli/gates/<gate-id>.json` store, and CLI commands for listing, showing, approving, and rejecting local gate decisions. This is local-only: no coordinator, network sync, daemon join, MCP, or remote execution behavior is introduced.

## Why

Gates are decision boundaries in execution contracts. Before Sykli can sync gate approvals through a team coordinator, it needs honest local decision state that can be persisted, inspected, and controlled. Manual gate decisions must not be silently overwritten or treated as successful without an explicit approval/rejection record.

## What changed

- Added `Sykli.GateDecision` for versioned local gate decision records.
- Added `Sykli.Gate.Store` backed by `.sykli/gates/<gate-id>.json`.
- Added strict gate status validation and transition rules:
  - `waiting` / `blocked` may transition to `approved` or `rejected`
  - `approved`, `rejected`, and `expired` are terminal
- Added `sykli gates list [--status STATUS]`.
- Added `sykli gate show <gate-id>`.
- Added `sykli gate approve <gate-id> --reason ...`.
- Added `sykli gate reject <gate-id> --reason ...`.
- Added `--json` support for every gate command.
- Added structured errors for missing gates, invalid IDs, malformed JSON, missing reasons, invalid decisions, and invalid transitions.
- Updated docs and README to document local gate state, commands, JSON shape, transition semantics, and public error codes.

## Runtime integration status

Runtime gate integration is deferred. This PR intentionally adds the local decision model/store/CLI only.

The existing executor gate path does not yet have a clean local state hook carrying the full run/work/gate lifecycle context needed to persist `.sykli/gates/<gate-id>.json` without broad executor changes. Existing runtime gates still do not silently pass; this PR does not weaken that behavior. The follow-up should thread gate request creation through execution once run/work context is available in the gate lifecycle.

## Validation

- [x] `python3 -m json.tool test/blackbox/dataset.json >/dev/null` passed
- [x] `mix format --check-formatted` passed
- [x] `git diff --check` passed
- [x] `mix test test/sykli/gate_decision_test.exs test/sykli/gate/store_test.exs test/sykli/cli/gate_test.exs test/sykli/cli_test.exs` passed: 56 tests, 0 failures
- [x] `mix escript.build` passed
- [x] `eval/oracle/run.sh --case 063` passed
- [x] `eval/oracle/run.sh --case 064` passed
- [x] `eval/oracle/run.sh --case 065` passed
- [x] `eval/oracle/run.sh --case 066` passed
- [x] `eval/oracle/run.sh --case 067` passed
- [x] `eval/oracle/run.sh --case 068` passed
- [x] `eval/oracle/run.sh --case 069` passed
- [x] `test/blackbox/run.sh --filter=GATE` passed: 3 passed, 0 failed, 155 skipped
- [x] `mix test` passed: 1 property, 1606 tests, 0 failures, 1 skipped, 13 excluded

Full conformance was not run because this PR does not change contract schemas, SDK output, or conformance fixtures.

## Oracle coverage

Added oracle cases:

- 063: gates list JSON returns local gate decisions
- 064: gate show JSON returns one gate
- 065: gate approve JSON records decision
- 066: gate reject JSON records decision
- 067: missing gate returns JSON error
- 068: invalid transition returns JSON error
- 069: invalid gate id returns JSON error

## Blackbox coverage

Added `GATE` blackbox category and scenarios:

- `GATE-001`: local gate decision CLI workflow persists state
- `GATE-002`: gate show rejects path traversal ids with JSON error
- `GATE-003`: terminal gate transition fails clearly

## Out of scope

- No coordinator
- No networking
- No daemon join
- No team sync
- No run summary sync to coordinator
- No gate approval sync through coordinator
- No MCP
- No Kubernetes deployment
- No remote execution
- No web UI
- No policy engine

## Next PR

Add self-hosted coordinator skeleton and store:

- `Sykli.Coordinator.Application`
- `Sykli.Coordinator.Router`
- `Sykli.Coordinator.Store`
- health endpoint
- org/team endpoints
- work item endpoints
- token auth
- coordinator API tests
- oracle JSON expectations
- blackbox service workflow if supported

### DIFF
--- a/core/lib/sykli/cli.ex
+++ b/core/lib/sykli/cli.ex
@@ -106,11 +106,17 @@ defmodule Sykli.CLI do
 
   def normalize_global_json(args), do: args
 
-  defp normalize_gates_args([]), do: ["list"]
-  defp normalize_gates_args(["list" | _] = args), do: args
+  @doc false
+  def normalize_gates_args([]), do: ["list"]
 
-  defp normalize_gates_args(args) do
-    if "list" in args, do: args, else: ["list" | args]
+  def normalize_gates_args([first | _] = args)
+      when first in ~w(list show approve reject --help -h),
+      do: args
+
+  def normalize_gates_args([<<"--", _::binary>> | _] = args), do: ["list" | args]
+
+  def normalize_gates_args(args) do
+    args
   end
 
   defp print_help do
--- a/core/lib/sykli/cli.ex
+++ b/core/lib/sykli/cli.ex
@@ -10,7 +10,7 @@ defmodule Sykli.CLI do
   alias Sykli.Work.Store, as: WorkStore
 
   @version Mix.Project.config()[:version]
-  @subcommands ~w(cache graph delta watch report history daemon work init validate verify plan explain context fix query mcp run)
+  @subcommands ~w(cache graph delta watch report history daemon work gates gate init validate verify plan explain context fix query mcp run)
 
   def main(args \\ []) do
     args = normalize_global_json(args)
@@ -49,6 +49,17 @@ defmodule Sykli.CLI do
       ["work" | work_args] ->
         work_args
         |> Sykli.CLI.Work.run()
+        |> halt()
+
+      ["gates" | gate_args] ->
+        gate_args
+        |> normalize_gates_args()
+        |> Sykli.CLI.Gate.run()
+        |> halt()
+
+      ["gate" | gate_args] ->
+        gate_args
+        |> Sykli.CLI.Gate.run()
         |> halt()
 
       ["init" | init_args] ->
@@ -95,6 +106,13 @@ defmodule Sykli.CLI do
 
   def normalize_global_json(args), do: args
 
+  defp normalize_gates_args([]), do: ["list"]
+  defp normalize_gates_args(["list" | _] = args), do: args
+
+  defp normalize_gates_args(args) do
+    if "list" in args, do: args, else: ["list" | args]
+  end
+
   defp print_help do
     IO.puts("""
     sykli - CI pipelines in your language
@@ -136,6 +154,8 @@ defmodule Sykli.CLI do
       sykli history    List recent runs
       sykli daemon     Manage daemon (see: sykli daemon --help)
       sykli work       Manage local work items (see: sykli work --help)
+      sykli gates      List local gate decisions
+      sykli gate       Manage local gate decisions (see: sykli gate --help)
       sykli cache      Manage cache (see: sykli cache --help)
       sykli mcp        Start MCP server (for AI assistants)
 
--- a/core/lib/sykli/cli/gate.ex
+++ b/core/lib/sykli/cli/gate.ex
@@ -86,12 +86,12 @@ defmodule Sykli.CLI.Gate do
 
   defp decide(action, id, opts, runtime_opts) do
     reason = Keyword.get(opts, :reason)
-    {_actor_type, actor_id} = actor(opts, runtime_opts)
+    {actor_type, actor_id} = actor(opts, runtime_opts)
 
     decision_opts =
       runtime_opts
       |> Keyword.take([:path, :now])
-      |> Keyword.put(:decided_by, actor_id)
+      |> Keyword.put(:decided_by, "#{actor_type}:#{actor_id}")
 
     result =
       case action do
--- a/core/lib/sykli/cli/gate.ex
+++ b/core/lib/sykli/cli/gate.ex
@@ -1,0 +1,334 @@
+defmodule Sykli.CLI.Gate do
+  @moduledoc """
+  Local gate decision CLI commands.
+
+  This module is intentionally local-only. Coordinator gate sync is added in a
+  later Team Mode phase.
+  """
+
+  alias Sykli.CLI.JsonResponse
+  alias Sykli.Error
+  alias Sykli.Error.Formatter
+  alias Sykli.Gate.Store
+  alias Sykli.GateDecision
+
+  @default_actor_type "member"
+
+  @doc """
+  Runs a `sykli gate ...` or `sykli gates ...` command and returns the intended
+  process exit code.
+
+  `runtime_opts` are internal test hooks for store path, clocks, ids, and
+  default actor attribution. Parsed CLI flags are kept separate in `opts`.
+  """
+  def run(args, runtime_opts \\ []) do
+    case parse(args) do
+      {:ok, command, opts} ->
+        execute(command, opts, runtime_opts)
+
+      {:error, reason, opts} ->
+        output_error(reason, Keyword.get(opts, :json, false))
+    end
+  end
+
+  defp execute(:list, opts, runtime_opts) do
+    store_opts = store_opts(runtime_opts)
+
+    result =
+      case Keyword.get(opts, :status) do
+        nil -> Store.list(store_opts)
+        status -> Store.list_by_status(status, store_opts)
+      end
+
+    case result do
+      {:ok, gates} ->
+        output_success(%{source: "local", gates: Enum.map(gates, &gate_map/1)}, opts, fn ->
+          if gates == [] do
+            IO.puts("No gates")
+          else
+            IO.puts("Gates:")
+
+            Enum.each(gates, fn gate ->
+              IO.puts("  #{gate.id}  #{gate.status}  #{gate.node_id || "-"}")
+            end)
+          end
+        end)
+
+      {:error, reason} ->
+        output_error(reason, Keyword.get(opts, :json, false))
+    end
+  end
+
+  defp execute({:show, id}, opts, runtime_opts) do
+    case Store.get(id, store_opts(runtime_opts)) do
+      {:ok, gate} ->
+        output_success(%{source: "local", gate: gate_map(gate)}, opts, fn ->
+          print_gate(gate)
+        end)
+
+      {:error, reason} ->
+        output_error(reason, Keyword.get(opts, :json, false))
+    end
+  end
+
+  defp execute({:approve, id}, opts, runtime_opts) do
+    decide(:approve, id, opts, runtime_opts)
+  end
+
+  defp execute({:reject, id}, opts, runtime_opts) do
+    decide(:reject, id, opts, runtime_opts)
+  end
+
+  defp execute(:help, _opts, _runtime_opts) do
+    print_help()
+    0
+  end
+
+  defp decide(action, id, opts, runtime_opts) do
+    reason = Keyword.get(opts, :reason)
+    {_actor_type, actor_id} = actor(opts, runtime_opts)
+
+    decision_opts =
+      runtime_opts
+      |> Keyword.take([:path, :now])
+      |> Keyword.put(:decided_by, actor_id)
+
+    result =
+      case action do
+        :approve -> Store.approve(id, reason, decision_opts)
+        :reject -> Store.reject(id, reason, decision_opts)
+      end
+
+    case result do
+      {:ok, gate} ->
+        verb = if action == :approve, do: "Approved", else: "Rejected"
+
+        output_success(%{source: "local", gate: gate_map(gate)}, opts, fn ->
+          IO.puts("#{verb} gate #{gate.id}")
+        end)
+
+      {:error, reason} ->
+        output_error(reason, Keyword.get(opts, :json, false))
+    end
+  end
+
+  defp parse(args) do
+    {opts, positionals, error} = parse_flags(args, [], [])
+
+    cond do
+      error != nil ->
+        {:error, error, opts}
+
+      positionals in [[], ["--help"], ["-h"]] ->
+        {:ok, :help, opts}
+
+      positionals == ["list"] ->
+        {:ok, :list, opts}
+
+      match?(["show", _], positionals) ->
+        ["show", id] = positionals
+        {:ok, {:show, id}, opts}
+
+      positionals == ["show"] ->
+        {:error, {:invalid_gate_command, "show requires a gate id"}, opts}
+
+      match?(["approve", _], positionals) ->
+        ["approve", id] = positionals
+        {:ok, {:approve, id}, opts}
+
+      positionals == ["approve"] ->
+        {:error, {:invalid_gate_command, "approve requires a gate id"}, opts}
+
+      match?(["reject", _], positionals) ->
+        ["reject", id] = positionals
+        {:ok, {:reject, id}, opts}
+
+      positionals == ["reject"] ->
+        {:error, {:invalid_gate_command, "reject requires a gate id"}, opts}
+
+      true ->
+        {:error, {:invalid_gate_command, Enum.join(positionals, " ")}, opts}
+    end
+  end
+
+  defp parse_flags([], opts, positionals),
+    do: {Enum.reverse(opts), Enum.reverse(positionals), nil}
+
+  defp parse_flags(["--json" | rest], opts, positionals) do
+    parse_flags(rest, [{:json, true} | opts], positionals)
+  end
+
+  defp parse_flags(["--help" | rest], opts, positionals) do
+    parse_flags(rest, opts, ["--help" | positionals])
+  end
+
+  defp parse_flags(["-h" | rest], opts, positionals) do
+    parse_flags(rest, opts, ["-h" | positionals])
+  end
+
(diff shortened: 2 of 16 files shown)


core/lib/sykli/cli.ex
```diff
@@ -95,6 +106,13 @@ defmodule Sykli.CLI do
 
   def normalize_global_json(args), do: args
 
+  defp normalize_gates_args([]), do: ["list"]
+  defp normalize_gates_args(["list" | _] = args), do: args
+
+  defp normalize_gates_args(args) do
+    if "list" in args, do: args, else: ["list" | args]
```
yairfalse: **Bug: `sykli gates show <id>` is corrupted by this normalizer.**

Trace:

- User types `sykli gates show gate_001`.
- Dispatch hits the `["gates" | gate_args]` branch with `gate_args = ["show", "gate_001"]`.
- `normalize_gates_args(["show", "gate_001"])` falls through to this clause: `"list" in args` is `false` → returns `["list", "show", "gate_001"]`.
- `Gate.run(["list", "show", "gate_001"])` → parser sees positionals `["list", "show", "gate_001"]` → no clause matches → `{:error, {:invalid_gate_command, "list show gate_001"}, opts}`.
- User gets: `Error: invalid local gate decision: invalid gate command "list show gate_001"`.

A plausibly typed command (English plural) becomes a confusing error message.

The normalizer should only insert `"list"` when there is no subcommand at all (just flags or empty args). Suggested fix:

```elixir
defp normalize_gates_args([]), do: ["list"]

defp normalize_gates_args([first | _] = args)
     when first in ~w(list show approve reject --help -h),
     do: args

defp normalize_gates_args([<<"--", _::binary>> | _] = args), do: ["list" | args]

defp normalize_gates_args(args), do: args  # let the parser produce a clear error
```

Add a test in `cli_test.exs` (or `cli/gate_test.exs`) for `sykli gates show gate_001` returning the gate, plus `sykli gates --status=waiting` defaulting to `list`.

core/lib/sykli/cli/gate.ex
```diff
@@ -0,0 +1,334 @@
...
+  end
+
+  defp decide(action, id, opts, runtime_opts) do
+    reason = Keyword.get(opts, :reason)
+    {_actor_type, actor_id} = actor(opts, runtime_opts)
+
+    decision_opts =
+      runtime_opts
+      |> Keyword.take([:path, :now])
+      |> Keyword.put(:decided_by, actor_id)
```
yairfalse: **`decided_by` loses actor type.**

```elixir
{_actor_type, actor_id} = actor(opts, runtime_opts)
...
|> Keyword.put(:decided_by, actor_id)
```

Gate creation captures both `requested_by_type` and `requested_by_id`. Decisions only capture `decided_by` (a single string), discarding the type. The design doc (`docs/self-hosted-coordinator.md`'s `gates` table) defines `decided_by` as a single field, so the implementation matches the doc — but in Phase 6 sync, the coordinator can't tell `--actor agent:claude` apart from `--actor member:claude`.

Two paths:

1. **Split locally**: add `decided_by_type` to `GateDecision`. Mirrors `requested_by_*` and `created_by_*` on work items. Requires a design-doc update to add the second column.
2. **Colon-string**: persist `decided_by: "agent:claude"`. Compact and parseable, matches the `--actor` input format. The design doc would just clarify the format.

I'd lean (1) for symmetry with the request side, but (2) requires no extra storage. Either is better than silently losing the type.

At minimum, document this in `Sykli.GateDecision`'s moduledoc so future contributors know `decided_by` is identity-only by design.

core/lib/sykli/gate_decision.ex
```diff
@@ -0,0 +1,257 @@
...
+    if value == "", do: nil, else: value
+  end
+
+  defp blank_to_nil(value), do: value
+
+  defp now_iso8601 do
+    DateTime.utc_now()
+    |> DateTime.truncate(:second)
+    |> DateTime.to_iso8601()
+  end
```
yairfalse, line 284: **Recurring pattern: third private `now_iso8601/0` in the codebase.**

`Sykli.WorkItem` and the run-history layer each have their own copy of this exact function, all calling `DateTime.utc_now()` directly. CLAUDE.md flags `DateTime.utc_now` as a `CredoSykli.NoWallClock` violation in simulator-facing code (the previous PRs apparently passed Credo, but `mix credo --strict` isn't in this PR's validation list — worth confirming).

Independent of whether Credo flags it, the duplication is a smell:

- Three places to update if the format ever changes.
- Three `now:` opt injection seams to keep consistent.
- Easy for a future contributor to copy a stale version.

Follow-up suggestion: extract `Sykli.Clock.now_iso8601/0` (or behaviour-split if tests need it) and have all three modules call it. Not blocking for this PR — but worth a tracking issue before Phase 4 (coordinator) introduces a fourth copy.

core/lib/sykli/gate_decision.ex
```diff
@@ -0,0 +1,257 @@
...
+  defp validate_requester_id(id), do: {:error, {:invalid_gate_requester_id, id}}
+
+  defp validate_evidence_refs(refs) when is_list(refs) do
+    Enum.reduce_while(refs, :ok, fn
+      ref, :ok when is_map(ref) -> {:cont, :ok}
+      ref, :ok -> {:halt, {:error, {:invalid_gate_evidence_ref, ref}}}
+    end)
+  end
+
+  defp validate_evidence_refs(refs), do: {:error, {:invalid_gate_evidence_refs, refs}}
```
yairfalse, line 264: **`evidence_refs` shape is unvalidated beyond list-of-maps.**

```elixir
defp validate_evidence_refs(refs) when is_list(refs) do
  Enum.reduce_while(refs, :ok, fn
    ref, :ok when is_map(ref) -> {:cont, :ok}
    ref, :ok -> {:halt, {:error, {:invalid_gate_evidence_ref, ref}}}
  end)
end
```

A persisted gate with `evidence_refs: [%{}]` or `evidence_refs: [%{"foo": "bar"}]` round-trips cleanly. The design doc (`docs/self-hosted-coordinator.md` `evidence_refs` table) calls for `type`, `uri`, `hash`, `summary`, `visibility` — none enforced here.

For Phase 3 this is acceptable since the only producers will be tests and (eventually) the executor. But future consumers (MCP tools surfacing evidence, `sykli explain` rendering it, Phase 7 sync deconstructing it into coordinator rows) will need to handle malformed refs defensively.

Two low-cost mitigations:

1. Tighten validation to require `"type"` and `"uri"` keys at minimum.
2. Add a moduledoc note on `Sykli.GateDecision`: "`evidence_refs` is stored as an opaque list of maps; consumers should treat unknown keys conservatively and validate the inner shape they need."

(2) is the lowest-friction path.

core/lib/sykli/gate_decision.ex
```diff
@@ -0,0 +1,257 @@
...
+  Local gate decision model for Team Mode.
+
+  A gate decision is a local coordination record stored under
+  `.sykli/gates/<id>.json`. It records a waiting or blocked decision boundary and
+  the later approval or rejection. Coordinator sync and runtime gate hookup are
+  later Team Mode layers.
+  """
+
+  @version "1"
+  @statuses ~w(waiting approved rejected blocked expired)
```
yairfalse, line 19: **No public API to transition a gate into `blocked` or `expired`.**

The state machine has five states but only two transition primitives (`approve/3` and `reject/3`). Today there's no way to:

- Mark a `waiting` gate as `blocked` (e.g., when a dependency fails mid-run).
- Mark a `waiting`/`blocked` gate as `expired` (e.g., from a future expiration sweeper).

New gates can be created in any valid status via `Store.create(status: ...)`, but existing gates can only be approved or rejected.

This is fine for Phase 3 (CLI exposes only approve/reject) but the runtime-hookup PR will need at least a `block/2` API, and any expiration job will need `expire/2`. Not blocking for this slice; worth a `# TODO: block/expire transitions for runtime + expirer` comment, or a tracking issue, so the next PR doesn't reinvent this.

yairfalse (commented):
## Overview

Local gate decision layer: `Sykli.GateDecision` (versioned model with status state machine), `Sykli.Gate.Store` (`.sykli/gates/<id>.json`), and `Sykli.CLI.Gate` (`gates list`, `gate show/approve/reject` with `--json`). Six new public-unstable error codes. Seven oracle cases (063–069) and three blackbox cases (`GATE-001/002/003`) including a terminal-transition negative.

The **PR description is exceptionally honest** about scope: it explicitly notes that runtime gate integration is deferred and that this PR ships only the decision model/store/CLI. The roadmap update reflects that, dropping the original "a blocked run resumes after `gate approve`" exit criterion to a follow-up. Honest scoping helps a lot — readers know what they're getting.

Things that look good:

- **Strict transition rules.** `waiting`/`blocked` → `approved`/`rejected`; `approved`/`rejected`/`expired` are terminal. Tests directly assert the negative path. `GATE-003` blackbox case verifies terminal gates can't be silently overwritten.
- **Reason is required** for approve/reject. Empty (`""`/whitespace) is rejected with `gate_decision_missing_reason`. Decisions can't slip through with no audit trail.
- **Data minimization assertion** in `GATE-001`: `! grep -q "secret" .sykli/gates/gate_001.json`. Same pattern as `WORK-003` — guards the local-state-plane boundary.
- **`malformed_gate_json` and missing-version errors** propagate cleanly through the CLI, with structurally-invalid persisted JSON yielding `{:missing_gate_version, nil}` rather than a confusing crash.
- **Reuses `Sykli.WorkItem.validate_id/1`** for `work_item_id` in gate decisions, so traversal-like values are rejected at the boundary.

One real bug, one design-consistency item, plus a few smaller things. See inline comments.

## Top concerns

### 1. `normalize_gates_args/1` corrupts `sykli gates show <id>`

The normalizer rewrites `gates show gate_001` into `["list", "show", "gate_001"]`, which the parser then rejects as `invalid local gate decision: invalid gate command "list show gate_001"`. The intended behavior is that `sykli gates` and `sykli gates --json` should default to `list`, but every other subcommand under `gates` should pass through unchanged. Fix in inline comment.

### 2. `decided_by` carries actor_id only — actor_type is discarded

The CLI captures `--actor TYPE:ID` for validation but only the id flows into `decided_by`. So `--actor agent:claude` and `--actor member:claude` are indistinguishable in the persisted gate file. The design doc has `decided_by` as a single field, so the implementation matches the doc — but in Phase 6 sync the coordinator won't know whether a decision came from a human, an agent, or a daemon. Worth either splitting (`decided_by_type`, `decided_by_id`) or adopting the colon-string convention. See inline comment.

### 3. `mix credo --strict` not in the validation list

Same as PR #189. Per project convention (`docs/done.md`, CLAUDE.md), and given the recurring `DateTime.utc_now/0` pattern in Phase 1/2/3 modules, worth confirming `mix credo --strict` passes against the `CredoSykli.NoWallClock` check.

## Recurring pattern: private `now_iso8601/0`

This is now the third Team Mode module with its own private `now_iso8601/0` (after `Sykli.WorkItem` and run-history changes). Each one calls `DateTime.utc_now()` directly. Even if Credo passes, the duplication is a smell:

- Three places to update if the timestamp format ever changes.
- Three injection seams (`now:` opt) that need to stay consistent.
- Three places that could drift if the next contributor copies an outdated one.

A shared `Sykli.Clock` module (with a behaviour for testability or just a function returning the canonical iso8601 string) would centralize this. Not blocking for Phase 3, but worth a tracking issue.

## Project convention check

| Rule | Status |
|---|---|
| `source: "local"` marker per design | ✓ |
| Path containment via `String.starts_with?(path, base <> "/")` | ✓ |
| Atomic file writes | ✓ |
| Closed-enum vocabulary modules | ✓ (`statuses/0`, `requester_types/0`) |
| Strict version-policy mirror of the engine | ✓ (`missing_gate_version` returned, no silent default) |
| Error codes added to `docs/error-codes.md` with line numbers | ✓ |
| `mix format --check-formatted` | ✓ (in PR description) |
| `mix test` | ✓ (1606 tests, 0 failures) |
| `mix credo --strict` | **not reported** — see top concern #3 |
| `git diff --check` | ✓ |
| Roadmap honest about deferred scope | ✓ |

## Test coverage

Good. Notable gaps:

- **`normalize_gates_args/1` corner cases** — no test exercises `sykli gates show <id>`, which surfaces issue #1.
- **`--reason ""` (empty after trim) at the CLI level** — `validate_decision_reason/1` rejects it, but the CLI test only covers missing `--reason` (which produces `nil`). Worth adding `--reason "   "` to verify trim handling.
- **Env-var fallback chain for `default_actor_id`** — same as PR 188; only the injection seam is tested, not the production fallback order.
- **State machine API gap** — no test for transitioning a gate into `expired` or `blocked` via decision API, because no such API exists. Future expirer/runtime will need it.

The test `"rejects invalid ids, versions, statuses, requesters, and evidence refs"` in `gate_decision_test.exs` is a great single-test sweep across multiple validators. Good consolidation.

## Security

- Gate id validation: `^[A-Za-z0-9][A-Za-z0-9_.-]{0,127}$`. Same regex as work items. Rejects `..`, `/`. ✓
- Path containment via `Path.expand` + `String.starts_with?`. ✓
- `evidence_refs` only validated as list-of-maps — no inner shape check. Acceptable for v0; worth a moduledoc note for future consumers (see inline comment).
- The `! grep -q "secret"` blackbox assertion is a soft regression guard. ✓
- `decided_by`/`decided_at` validated as optional strings only — no format enforcement. Acceptable since they're not used as path components.

No new attack surface beyond Phases 1/2.

## Performance

- `list/1` is O(N) over all gate files. Fine for typical Phase 3 quantities.
- `list_by_status/2` filters after `list/1` — same after-limit pattern as `RunHistory.list_by_work_item/2` from PR #189. Less of a concern here since gate files are typically few; no `:limit` opt is exposed.
- Atomic write per call. No batched I/O.

## Bottom line

Well-scoped half of Phase 3 — local decision model + store + CLI, with the runtime hookup honestly deferred. The `normalize_gates_args` bug is the only thing I'd want fixed before merge; the `decided_by` attribution and shared-clock items are best resolved as follow-ups before Phase 6 sync. Test coverage is strong and the data-minimization guards are well-placed.

LGTM with #1 fixed.

core/lib/sykli/cli.ex
yairfalse: Fixed in 301f5a1. `normalize_gates_args/1` now only inserts `list` for empty/flag-only plural `gates` invocations, leaves explicit subcommands (`show`, `approve`, `reject`) untouched, and lets unknown subcommands flow to the normal parser error. Added coverage for `gates --status=waiting`, `gates show gate_001`, explicit approve, and unknown plural subcommands.

core/lib/sykli/cli/gate.ex
yairfalse: Fixed in 301f5a1 using the compact colon-string path. The CLI now persists `decided_by` as `TYPE:ID`, e.g. `agent:claude`, so actor type is not lost without changing the v0 field shape. Added CLI test coverage for `--actor=agent:claude` and documented the format in `Sykli.GateDecision` and `docs/local-state-plane.md`.

core/lib/sykli/gate_decision.ex
yairfalse: Confirmed with validation: `mix credo --strict` passes on this branch. I did not extract a shared clock in this PR because it would broaden this local gate slice across existing WorkItem/run-history code, but the concern is valid for the next coordinator/store layer before another timestamp helper is added.

core/lib/sykli/gate_decision.ex
yairfalse: Addressed in 301f5a1 with the low-friction path: `Sykli.GateDecision` now documents `evidence_refs` as opaque local v0 maps and requires consumers to validate the inner shape they need before rendering or syncing. I kept the list-of-maps storage permissive for this slice so future executor/coordinator producers can evolve without forcing a schema change here.

core/lib/sykli/gate_decision.ex
yairfalse: Fixed in 301f5a1. Added small `GateDecision.block/2` and `GateDecision.expire/2` APIs plus tests. They are not exposed in CLI yet, but the next runtime hook/expirer layer now has explicit transition primitives instead of having to mutate status directly.

PR #191, opened 2026-05-09
Add self-hosted coordinator skeleton

yairfalse:
## Summary

This PR adds the first self-hosted Team Mode coordinator skeleton. It introduces an opt-in coordinator HTTP API with health, org, team, and work item endpoints, plus minimal bearer-token auth for all non-health endpoints.

This does not add daemon join, heartbeat, team sync, run sync, gate sync, MCP tools, Kubernetes deployment, or remote execution.

## Why

Local Team Mode now has work items, run association, and local gate decision state. Teams need a shared coordination point for org/team work metadata before daemons can join or sync run/gate projections.

The coordinator is for shared state, not execution. The existing BEAM mesh remains the trusted execution-domain mechanism; this coordinator starts the remote/team coordination plane safely.

## What changed

- Added `Sykli.Coordinator.Application` as an opt-in Bandit supervisor.
- Added `Sykli.Coordinator.Router` with JSON envelope responses.
- Added `Sykli.Coordinator.Auth` for minimal bearer-token auth.
- Added `Sykli.Coordinator.Store` as an in-memory skeleton store.
- Added `sykli coordinator start --token TOKEN [--port PORT]`.
- Added endpoints:
  - `GET /health`
  - `GET /healthz`
  - `POST /v1/orgs`
  - `GET /v1/orgs`
  - `POST /v1/teams`
  - `GET /v1/teams`
  - `GET /v1/work-items`
  - `POST /v1/work-items`
  - `GET /v1/work-items/:id`
  - `POST /v1/work-items/:id/claim`
  - `POST /v1/work-items/:id/notes`
- Added structured coordinator API/CLI error codes.
- Added focused store/router/auth tests.
- Added coordinator oracle JSON cases.
- Added a blackbox coordinator service workflow.
- Updated Team Mode docs and error-code catalog.

## Store choice and limitations

This PR implements an in-memory coordinator store. It is intentionally not production persistence.

The in-memory store is sufficient for this skeleton PR because it locks in the API shape, auth boundary, JSON envelopes, and test harness without pulling Postgres migrations into the first networked slice. Durable Postgres storage and migrations remain follow-up work.

## Security behavior

- `GET /health` and `GET /healthz` are unauthenticated.
- All `/v1/*` endpoints require `Authorization: Bearer <token>`.
- Missing, malformed, or incorrect tokens are rejected.
- No remote execution endpoints exist.
- No daemon session endpoints exist.
- No run/gate/evidence sync endpoints exist.
- No log, artifact, source code, or raw stdout/stderr upload endpoints exist.
- No anonymous writes.

## Validation

- [x] `cd core && mix format` passed.
- [x] `cd core && mix test test/sykli/coordinator/store_test.exs test/sykli/coordinator/auth_test.exs test/sykli/coordinator/router_test.exs` passed: 13 tests, 0 failures.
- [x] `cd core && mix escript.build` passed.
- [x] `test/blackbox/run.sh --filter=COORD-001` passed: 1 passed, 0 failed, 158 skipped.
- [x] `eval/oracle/run.sh --case 070 && eval/oracle/run.sh --case 071 && eval/oracle/run.sh --case 072 && eval/oracle/run.sh --case 073` passed: all 4 coordinator oracle cases passed.
- [x] `cd core && mix credo --strict` passed: found no issues.
- [x] `cd core && mix test` passed: 1 property, 1623 tests, 0 failures, 1 skipped, 13 excluded.
- [x] `git diff --check` passed.

Full conformance was not run because this PR does not change contract schemas, SDK emitters, or conformance fixtures.

## Oracle coverage

Added cases:

- `070` coordinator health JSON envelope
- `071` coordinator authenticated work API JSON
- `072` coordinator unauthorized JSON error
- `073` coordinator duplicate org slug JSON error

## Blackbox coverage

Added `COORD-001`, which starts the coordinator service, verifies health, verifies unauthorized write rejection, creates org/team/work item state, claims a work item, appends a note, and verifies the listed work item is a safe shared projection without logs/artifacts/secrets.

## Out of scope

- No daemon join.
- No heartbeat.
- No team sync from local CLI.
- No run sync.
- No gate sync.
- No MCP tools.
- No Kubernetes deployment.
- No remote execution.
- No remote shell.
- No web UI.
- No SaaS.
- No complex RBAC.
- No OIDC or GitHub org mapping.
- No artifact/log hosting.

## Next PR

Add daemon join and heartbeat sync:

- `POST /v1/daemon-sessions`
- `POST /v1/daemon-sessions/:id/heartbeat`
- daemon join CLI
- heartbeat payloads
- `accepts_remote_work` default false
- daemon session tests
- oracle/blackbox coverage

### DIFF
--- a/core/lib/sykli/cli/coordinator.ex
+++ b/core/lib/sykli/cli/coordinator.ex
@@ -11,6 +11,7 @@ defmodule Sykli.CLI.Coordinator do
   alias Sykli.Error.Formatter
 
   @default_port 8620
+  @default_bind "127.0.0.1"
 
   def run(args) do
     case parse(args) do
@@ -31,19 +32,27 @@ defmodule Sykli.CLI.Coordinator do
 
     with {:ok, token} <- token(opts),
          {:ok, port} <- port(opts),
-         {:ok, _pid} <- Sykli.Coordinator.Application.start_link(token: token, port: port) do
+         {:ok, bind_string, bind_address} <- bind(opts),
+         {:ok, _pid} <-
+           Sykli.TeamCoordinator.Application.start_link(
+             token: token,
+             port: port,
+             bind: bind_address
+           ) do
       if json? do
         IO.puts(
           JsonResponse.ok(%{
             source: "coordinator",
             status: "running",
+            bind: bind_string,
             port: port,
             storage: "in_memory"
           })
         )
       else
-        IO.puts("Sykli coordinator listening on http://127.0.0.1:#{port}")
+        IO.puts("Sykli coordinator listening on http://#{bind_string}:#{port}")
         IO.puts("storage: in_memory")
+        IO.puts("warning: plain HTTP; terminate TLS at an ingress or proxy for production")
       end
 
       Process.sleep(:infinity)
@@ -89,6 +98,12 @@ defmodule Sykli.CLI.Coordinator do
   defp parse_flags([<<"--port=", value::binary>> | rest], opts, positionals),
     do: parse_flags(rest, [{:port, value} | opts], positionals)
 
+  defp parse_flags(["--bind", value | rest], opts, positionals),
+    do: parse_flags(rest, [{:bind, value} | opts], positionals)
+
+  defp parse_flags([<<"--bind=", value::binary>> | rest], opts, positionals),
+    do: parse_flags(rest, [{:bind, value} | opts], positionals)
+
   defp parse_flags(["--token", value | rest], opts, positionals),
     do: parse_flags(rest, [{:token, value} | opts], positionals)
 
@@ -125,6 +140,21 @@ defmodule Sykli.CLI.Coordinator do
     end
   end
 
+  defp bind(opts) do
+    value = Keyword.get(opts, :bind, @default_bind)
+
+    with {:ok, address} <- parse_bind(value) do
+      {:ok, value, address}
+    end
+  end
+
+  defp parse_bind(value) when is_binary(value) do
+    case :inet.parse_address(String.to_charlist(value)) do
+      {:ok, address} -> {:ok, address}
+      {:error, _reason} -> {:error, {:invalid_coordinator_bind, value}}
+    end
+  end
+
   defp output_error(reason, json_output) do
     error = coordinator_error(reason)
 
@@ -154,6 +184,16 @@ defmodule Sykli.CLI.Coordinator do
       message: "invalid coordinator port: #{inspect(value)}",
       step: :setup,
       hints: ["use --port with a TCP port between 1 and 65535"]
+    }
+  end
+
+  defp coordinator_error({:invalid_coordinator_bind, value}) do
+    %Error{
+      code: "coordinator.invalid_bind",
+      type: :validation,
+      message: "invalid coordinator bind address: #{inspect(value)}",
+      step: :setup,
+      hints: ["use --bind with an IPv4 or IPv6 address, for example 127.0.0.1"]
     }
   end
 
@@ -205,12 +245,13 @@ defmodule Sykli.CLI.Coordinator do
     Self-hosted Team Mode coordinator commands.
 
     Commands:
-      sykli coordinator start --token TOKEN [--port PORT]
+      sykli coordinator start --token TOKEN [--port PORT] [--bind ADDRESS]
 
     Options:
       --json          Output startup status as JSON before serving
       --token TOKEN   Bearer token required for non-health API endpoints
       --port PORT     HTTP port (default: #{@default_port})
+      --bind ADDRESS  Listen address (default: #{@default_bind}; use 0.0.0.0 only intentionally)
 
     The coordinator skeleton stores state in memory. It exposes health,
     org/team, and work item endpoints only; it does not execute work.
--- a/core/lib/sykli/cli/coordinator.ex
+++ b/core/lib/sykli/cli/coordinator.ex
@@ -1,0 +1,219 @@
+defmodule Sykli.CLI.Coordinator do
+  @moduledoc """
+  Minimal CLI entry point for the self-hosted Team Mode coordinator skeleton.
+
+  This command only starts the coordinator HTTP API. It does not join daemons,
+  sync runs, sync gates, or execute remote work.
+  """
+
+  alias Sykli.CLI.JsonResponse
+  alias Sykli.Error
+  alias Sykli.Error.Formatter
+
+  @default_port 8620
+
+  def run(args) do
+    case parse(args) do
+      {:ok, :help, _opts} ->
+        print_help()
+        0
+
+      {:ok, :start, opts} ->
+        start(opts)
+
+      {:error, reason, opts} ->
+        output_error(reason, Keyword.get(opts, :json, false))
+    end
+  end
+
+  defp start(opts) do
+    json? = Keyword.get(opts, :json, false)
+
+    with {:ok, token} <- token(opts),
+         {:ok, port} <- port(opts),
+         {:ok, _pid} <- Sykli.Coordinator.Application.start_link(token: token, port: port) do
+      if json? do
+        IO.puts(
+          JsonResponse.ok(%{
+            source: "coordinator",
+            status: "running",
+            port: port,
+            storage: "in_memory"
+          })
+        )
+      else
+        IO.puts("Sykli coordinator listening on http://127.0.0.1:#{port}")
+        IO.puts("storage: in_memory")
+      end
+
+      Process.sleep(:infinity)
+      0
+    else
+      {:error, reason} -> output_error(reason, json?)
+    end
+  end
+
+  defp parse(args) do
+    {opts, positionals, error} = parse_flags(args, [], [])
+
+    cond do
+      error != nil ->
+        {:error, error, opts}
+
+      positionals in [[], ["--help"], ["-h"]] ->
+        {:ok, :help, opts}
+
+      positionals == ["start"] ->
+        {:ok, :start, opts}
+
+      true ->
+        {:error, {:invalid_coordinator_command, Enum.join(positionals, " ")}, opts}
+    end
+  end
+
+  defp parse_flags([], opts, positionals),
+    do: {Enum.reverse(opts), Enum.reverse(positionals), nil}
+
+  defp parse_flags(["--json" | rest], opts, positionals),
+    do: parse_flags(rest, [{:json, true} | opts], positionals)
+
+  defp parse_flags(["--help" | rest], opts, positionals),
+    do: parse_flags(rest, opts, ["--help" | positionals])
+
+  defp parse_flags(["-h" | rest], opts, positionals),
+    do: parse_flags(rest, opts, ["-h" | positionals])
+
+  defp parse_flags(["--port", value | rest], opts, positionals),
+    do: parse_flags(rest, [{:port, value} | opts], positionals)
+
+  defp parse_flags([<<"--port=", value::binary>> | rest], opts, positionals),
+    do: parse_flags(rest, [{:port, value} | opts], positionals)
+
+  defp parse_flags(["--token", value | rest], opts, positionals),
+    do: parse_flags(rest, [{:token, value} | opts], positionals)
+
+  defp parse_flags([<<"--token=", value::binary>> | rest], opts, positionals),
+    do: parse_flags(rest, [{:token, value} | opts], positionals)
+
+  defp parse_flags([<<"--", _::binary>> = flag | rest], opts, positionals) do
+    opts =
+      if "--json" in rest do
+        [{:json, true} | opts]
+      else
+        opts
+      end
+
+    {Enum.reverse(opts), Enum.reverse(positionals), {:unknown_coordinator_flag, flag}}
+  end
+
+  defp parse_flags([arg | rest], opts, positionals),
+    do: parse_flags(rest, opts, [arg | positionals])
+
+  defp token(opts) do
+    case Keyword.get(opts, :token) || System.get_env("SYKLI_COORDINATOR_TOKEN") do
+      value when is_binary(value) and value != "" -> {:ok, value}
+      _ -> {:error, :coordinator_token_required}
+    end
+  end
+
+  defp port(opts) do
+    value = Keyword.get(opts, :port, Integer.to_string(@default_port))
+
+    case Integer.parse(value) do
+      {port, ""} when port > 0 and port < 65_536 -> {:ok, port}
+      _ -> {:error, {:invalid_coordinator_port, value}}
+    end
+  end
+
+  defp output_error(reason, json_output) do
+    error = coordinator_error(reason)
+
+    if json_output do
+      IO.puts(JsonResponse.error(error))
+    else
+      IO.puts(:stderr, Formatter.format(error))
+    end
+
+    1
+  end
+
+  defp coordinator_error(:coordinator_token_required) do
+    %Error{
+      code: "coordinator.token_required",
+      type: :validation,
+      message: "coordinator start requires a token",
+      step: :setup,
+      hints: ["set SYKLI_COORDINATOR_TOKEN or pass --token <token>"]
+    }
+  end
+
+  defp coordinator_error({:invalid_coordinator_port, value}) do
+    %Error{
+      code: "coordinator.invalid_port",
+      type: :validation,
+      message: "invalid coordinator port: #{inspect(value)}",
+      step: :setup,
+      hints: ["use --port with a TCP port between 1 and 65535"]
+    }
+  end
+
+  defp coordinator_error({:unknown_coordinator_flag, flag}) do
+    %Error{
+      code: "coordinator.invalid_command",
+      type: :validation,
+      message: "unknown coordinator flag: #{flag}",
+      step: :validate,
+      hints: ["use: sykli coordinator start --token <token>"]
+    }
+  end
+
+  defp coordinator_error({:invalid_coordinator_command, command}) do
+    %Error{
+      code: "coordinator.invalid_command",
+      type: :validation,
+      message: "invalid coordinator command: #{command}",
+      step: :validate,
+      hints: ["use: sykli coordinator start --token <token>"]
+    }
+  end
+
+  defp coordinator_error({:shutdown, {:failed_to_start_child, _child, reason}}) do
+    %Error{
+      code: "coordinator.start_failed",
+      type: :runtime,
+      message: "coordinator failed to start",
+      step: :setup,
+      cause: reason,
+      hints: ["check whether the configured port is already in use"]
+    }
+  end
+
+  defp coordinator_error(reason) do
+    %Error{
+      code: "coordinator.start_failed",
+      type: :runtime,
+      message: "coordinator failed to start: #{inspect(reason)}",
+      step: :setup,
+      hints: []
+    }
+  end
+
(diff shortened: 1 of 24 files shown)


core/lib/sykli/team_coordinator/application.ex
```diff
@@ -0,0 +1,42 @@
...
+  def init(opts) do
+    token = Keyword.fetch!(opts, :token)
+    port = Keyword.get(opts, :port, @default_port)
+    store_name = Keyword.get(opts, :store_name, Sykli.Coordinator.Store)
+
+    children = [
+      {Sykli.Coordinator.Store, store_opts(opts, store_name)},
+      {Bandit,
+       plug: {Sykli.Coordinator.Router, store: store_name, token: token},
+       port: port,
```
yairfalse, line 33: **Bandit binds to `0.0.0.0` here, but the CLI claims `127.0.0.1`.**

```elixir
{Bandit,
 plug: {Sykli.Coordinator.Router, store: store_name, token: token},
 port: port,
 startup_log: false}
```

No `ip:` option → Bandit's default is `0.0.0.0` (all interfaces). The CLI in `cli/coordinator.ex:45` prints:

```elixir
IO.puts("Sykli coordinator listening on http://127.0.0.1:#{port}")
```

Misleading. A user on a laptop expects localhost-only; they get a service reachable from the LAN, protected only by the bearer token they configured.

Fix:

```elixir
{Bandit,
 plug: {Sykli.Coordinator.Router, store: store_name, token: token},
 ip: Keyword.get(opts, :bind, {127, 0, 0, 1}),
 port: port,
 startup_log: false}
```

And add a `--bind` flag (or `SYKLI_COORDINATOR_BIND`) for production behind Ingress. Update the CLI startup message to reflect the actual bind address.

This matters more than usual because the security doc emphasizes "network opt-in" — the current default is "network on, with one bearer token guarding the door."

core/lib/sykli/team_coordinator/application.ex
```diff
@@ -0,0 +1,42 @@
+defmodule Sykli.Coordinator.Application do
+  @moduledoc """
+  Supervisor for the self-hosted Team Mode coordinator skeleton.
+
+  This is opt-in and is not started by `Sykli.Application`. The existing
```
yairfalse, line 5: **`Sykli.Coordinator` namespace collision.**

```
This is opt-in and is not started by `Sykli.Application`. The existing
`Sykli.Coordinator` BEAM mesh process remains separate.
```

The moduledoc admits the issue but doesn't fix it. After this PR:

- `Sykli.Coordinator` — existing BEAM mesh coordinator role.
- `Sykli.Coordinator.Application` — Phase 4 HTTP service supervisor.
- `Sykli.Coordinator.Router` — Phase 4 Plug router.
- `Sykli.Coordinator.Store` — Phase 4 GenServer.
- `Sykli.Coordinator.Auth` — Phase 4 bearer auth.

The whole Team Mode design rests on the **mesh ≠ coordinator** distinction (`docs/coordination-modes.md`). The namespace doesn't reflect that.

Two cleaner options:

1. **Rename new modules to `Sykli.TeamCoordinator.*`.** Mirrors the doc vocabulary; explicitly distinguishes from the mesh role.
2. **Move the existing mesh role to `Sykli.Mesh.Coordinator`** and keep this tree under `Sykli.Coordinator.*`.

Either is a mechanical refactor today (5 new modules, ~2 hours of grep + sed + tests). Each subsequent Phase 5+ PR makes it harder.

I'd lean (1) — leaves the existing mesh role's name unchanged and gives the new HTTP service a name that clearly tells the reader which trust domain it's in.

core/lib/sykli/cli/coordinator.ex
```diff
@@ -0,0 +1,219 @@
...
+            port: port,
+            storage: "in_memory"
+          })
+        )
+      else
+        IO.puts("Sykli coordinator listening on http://127.0.0.1:#{port}")
+        IO.puts("storage: in_memory")
+      end
+
+      Process.sleep(:infinity)
```
yairfalse, line 58: **No graceful shutdown; SIGTERM kills BEAM hard.**

```elixir
Process.sleep(:infinity)
0
```

Common escript pattern, but `Sykli.Application` already installs a SIGTERM handler that drains in-flight `TaskSupervisor` children within `SYKLI_DRAIN_TIMEOUT_MS` (per CLAUDE.md). The coordinator skeleton has no equivalent — SIGTERM kills the BEAM, in-flight HTTP requests are dropped mid-response, audit rows for in-progress writes may never run.

For the in-memory skeleton this is acceptable (state is lost on restart anyway), but worth a `# TODO: hook a SIGTERM trap that calls Bandit's graceful shutdown` comment to mark the seam for the Postgres slice.

Also: the `0` return is unreachable. `Process.sleep(:infinity)` blocks forever, so neither the success exit code nor the `halt()` in `Sykli.CLI` ever runs. Cosmetic, but the dead `0` reads as "this returns 0 on success" when it actually never returns at all.

core/lib/sykli/team_coordinator/auth.ex
```diff
@@ -0,0 +1,46 @@
...
+    else
+      {:error, :missing_token_config} -> {:error, :coordinator_auth_not_configured}
+      {:error, reason} -> {:error, reason}
+      false -> {:error, :coordinator_unauthorized}
+    end
+  end
+
+  defp bearer_token(conn) do
+    case get_req_header(conn, "authorization") do
+      ["Bearer " <> token] when token != "" -> {:ok, token}
```
yairfalse, line 35: **Bearer scheme is case-sensitive — RFC 6750 says it shouldn't be.**

```elixir
["Bearer " <> token] when token != "" -> {:ok, token}
```

This matches `"Bearer "` exactly. RFC 6750 §2.1 specifies the scheme is case-insensitive (`Bearer` SHOULD be capitalized but `bearer`, `BEARER`, etc. are valid). Most clients use the canonical capitalization, but some (especially OpenAPI-generated clients in less common stacks, or curl `--header` typos) won't.

Low-friction fix:

```elixir
[<<scheme::binary-size(7), token::binary>>] when token != "" ->
  if String.downcase(scheme) == "bearer " do
    {:ok, token}
  else
    {:error, :coordinator_malformed_auth}
  end
```

Or, more cleanly, parse the header with `String.split(" ", parts: 2)` and downcase the scheme. Either way, add a test that asserts `"authorization: bearer X"` (lowercase) succeeds — currently no test covers this.

core/lib/sykli/team_coordinator/store.ex
```diff
@@ -0,0 +1,344 @@
...
+     %{
+       now: Keyword.get(opts, :now, &DateTime.utc_now/0),
+       id: Keyword.get(opts, :id, &Sykli.ULID.generate/0),
+       orgs: %{},
+       orgs_by_slug: %{},
+       teams: %{},
+       teams_by_org_slug: %{},
+       work_items: %{},
+       notes_by_work_item: %{},
+       audit_log: []
```
yairfalse, line 50: **Audit log grows without bound.**

```elixir
audit_log: []
```

Every state-changing call appends via `audit/6`. There's no rotation, no eviction, no max-size policy. For the in-memory skeleton this is acceptable (state is lost on restart), but the security doc says:

> The audit log is mandatory. … The coordinator never deletes audit log rows.

The current implementation:

- Writes audit rows ✓
- Never deletes them ✓ (in-memory)
- Loses everything on restart ✓ (acceptable per skeleton design)
- Grows BEAM heap unboundedly during long-running sessions — not yet a concern at low volume, but a future Postgres slice will need a documented retention policy

Two small additions worth landing:

1. Moduledoc note: "In-memory audit log is unbounded; expect O(API calls) memory growth. Postgres slice should land with a retention policy."
2. A `GET /v1/audit-log` endpoint (admin-only) so operators can inspect it without writing test code. Today the `audit_log/1` getter is reachable only from Elixir, not over HTTP.

Neither blocks merge.

core/lib/sykli/team_coordinator/router.ex
```diff
@@ -0,0 +1,239 @@
...
+         {:ok, org} <- Store.create_org(store(opts), body) do
+      send_json(conn, 201, %{org: org})
+    else
+      {:error, reason} -> send_error(conn, reason)
+    end
+  end
+
+  defp route_v1(%Plug.Conn{method: "GET", path_info: ["v1", "orgs"]} = conn, opts) do
+    with {:ok, orgs} <- Store.list_orgs(store(opts)) do
+      send_json(conn, 200, %{items: orgs})
```
yairfalse, line 47: **`with` without `else` on infallible-today calls.**

```elixir
defp route_v1(%Plug.Conn{method: "GET", path_info: ["v1", "orgs"]} = conn, opts) do
  with {:ok, orgs} <- Store.list_orgs(store(opts)) do
    send_json(conn, 200, %{items: orgs})
  end
end
```

`Store.list_orgs/1` only returns `{:ok, _}` today, so this works. But if a future refactor adds an error path (e.g., when the store is replaced by Postgres), the `with` will silently fall through without sending a response — leaving the HTTP connection hung until Bandit times it out.

Defensive shape:

```elixir
with {:ok, orgs} <- Store.list_orgs(store(opts)) do
  send_json(conn, 200, %{items: orgs})
else
  {:error, reason} -> send_error(conn, reason)
end
```

Same for `GET /v1/teams` (line 62) and `GET /v1/work-items` (line 75-78). Three lines per handler that prevent a future silent-hang regression.

yairfalse (commented):
## Overview

First server-side slice: a coordinator HTTP API with Bandit + Plug router, in-memory `GenServer` store, bearer-token auth on `/v1/*` endpoints, and a `sykli coordinator start` CLI. ~1.5k LoC across five new modules plus tests, oracle 070–073, and `COORD-001` blackbox. PR description is honest that this is a skeleton — Postgres, migrations, daemon sessions, runs, gates, and evidence are all explicitly deferred.

Things that look genuinely good:

- **Auth boundary is correct.** `route_v1` gates every `/v1/*` request through `Auth.authorize/2`; health is unauthenticated; `Plug.Crypto.secure_compare` provides constant-time matching with a length-mismatch fast-path.
- **"Does not expose dangerous execution endpoints" is a test, not a comment.** `router_test.exs` asserts `POST /v1/execute` returns 404. That's exactly the right kind of regression-prevention to land before more endpoints accumulate.
- **JSON envelope reuse.** Both the CLI and the HTTP API route through `Sykli.CLI.JsonResponse`, so agents see the same `ok/version/data/error` shape across surfaces.
- **Status codes are HTTP-shaped.** 401 for missing/wrong tokens, 404 for missing resources, 409 for duplicate slug/already-claimed, 413 for body too large. Sensible mapping at `status_for/1`.
- **Honest roadmap update.** `team-mode-roadmap.md` Phase 4 status now explicitly notes `migrate` and `status` subcommands are deferred along with Postgres.
- **Audit log writes** are consistent: every state-changing call invokes `audit/6`. `audit_log/1` getter is tested.

Two real concerns are worth fixing before more code stacks. See inline comments.

## Top concerns

### 1. `Sykli.Coordinator` namespace collision

The new `Sykli.Coordinator.{Application,Router,Store,Auth}` modules sit in the same top-level prefix as the existing `Sykli.Coordinator` (the BEAM mesh's coordinator role). The application moduledoc admits this directly:

> The existing `Sykli.Coordinator` BEAM mesh process remains separate.

This is a structural ambiguity: "coordinator" now refers to two different things in the codebase, depending on which submodule you're under. The whole Team Mode design rests on the **mesh ≠ coordinator** distinction (see `docs/coordination-modes.md`), and the namespace doesn't reflect it. Future contributors will conflate the two.

Fix now while only five modules use the prefix:

- `Sykli.TeamCoordinator.{Application,Router,Store,Auth}` — clear name, mirrors the doc vocabulary.
- Or: rename the existing mesh role module to `Sykli.Mesh.Coordinator` and keep the new tree under `Sykli.Coordinator.*`.

Both are mechanical refactors that get harder with each subsequent slice.

### 2. Bandit binds to 0.0.0.0; CLI says `http://127.0.0.1`

```elixir
{Bandit,
 plug: {Sykli.Coordinator.Router, store: store_name, token: token},
 port: port,
 startup_log: false}
```

No `ip:` option → Bandit's default is `0.0.0.0` (all interfaces). But the CLI prints:

```elixir
IO.puts("Sykli coordinator listening on http://127.0.0.1:#{port}")
```

Misleading, and a real concern given the messaging around "local-first, network opt-in." A user running `sykli coordinator start` on a laptop expects localhost; they get a service reachable from the LAN, protected only by the bearer token they configured. That's not catastrophic (token is mandatory), but it violates the principle of least surprise and the security doc's framing.

Fix: bind explicitly to `127.0.0.1` by default, add `--bind` for production deploys behind an Ingress, and update the message to reflect the actual binding.

## Recurring pattern: clock injection (now five modules)

`Sykli.WorkItem`, `Sykli.GateDecision`, `Sykli.RunHistory`, the work-item path in `cli.ex`, and now `Sykli.Coordinator.Store` all roll their own clock injection. PR #190 review flagged this; this PR adds a fifth instance. The shape varies (private `now_iso8601/0`, in-state function, ...) but the intent is identical.

Recommend extracting `Sykli.Clock` (callable: `Sykli.Clock.now_iso8601/0`; behaviour-injectable for tests) before Phase 5 lands a sixth copy.

## Project convention check

| Rule | Status |
|---|---|
| `--json` envelope via `JsonResponse` | ✓ (CLI + HTTP) |
| Constant-time secret comparison | ✓ (`Plug.Crypto.secure_compare`) |
| Closed-enum vocabulary modules | ✓ (`WorkItem.assignment_types/0`) |
| Error codes added to `docs/error-codes.md` with line numbers | ✓ |
| `mix format` | ✓ |
| `mix credo --strict` | ✓ (in PR description) |
| `mix test` | ✓ (1623 tests, 0 failures) |
| `git diff --check` | ✓ |
| Roadmap honest about deferred scope | ✓ |

## Test coverage

Good for the happy paths, auth boundary, and core errors. Gaps:

- **Lowercase `bearer` header.** `"Authorization: bearer secret"` (RFC 6750 says scheme is case-insensitive) currently fails. Either tighten to require uppercase `Bearer` and document, or accept any case.
- **Body too large** (>1MB). The router maps `:more` to `coordinator.payload_too_large` (413), but no test exercises this path.
- **Empty bearer.** `"Authorization: Bearer "` falls through to `[_]` → malformed_auth. Tested only indirectly via the malformed-token test.
- **Audit log endpoint coverage.** No HTTP route exposes the audit log; an operator can't inspect it in a running service. Phase 4 may not need it, but `GET /v1/audit-log` (admin-only) would be a small addition with high value.
- **Concurrent requests.** Single GenServer serializes everything; no test asserts behavior under parallel writes.
- **Coordinator over time** — audit log grows unbounded. State is in-memory so this disappears on restart, but a long-running skeleton accumulates audit rows forever. Not blocking for the skeleton; worth a moduledoc note.

The test `"does not expose dangerous execution endpoints"` (router_test.exs:147) is a model of what defensive Sykli testing should look like — strong contract assertion that future PRs can't accidentally erode.

## Security

- TLS termination is documented (Ingress recommended). The skeleton itself speaks plain HTTP, which is consistent with the design but should be louder in startup output (e.g., "WARNING: HTTP only — terminate TLS at Ingress for production").
- Bearer token mandatory on `/v1/*`. ✓
- No execution endpoints. ✓ (Tested.)
- No log/artifact/source upload endpoints. ✓
- 1MB body limit + 15s read timeout. ✓
- Audit log captures every state change. ✓ (in-memory, lost on restart per design.)
- Errors return generic messages — no internal stack traces leaked. ✓
- Token comparison constant-time. ✓
- Token stored in env var (`SYKLI_COORDINATOR_TOKEN`) per the security doc convention. ✓

The `0.0.0.0` binding (issue #2) is the only real attack-surface concern.

## Performance

- All API calls serialize through one GenServer. Acceptable for skeleton volumes; reads block writes. Worth a moduledoc note saying "all calls serialize; Postgres-backed migration should support read concurrency."
- Audit log unbounded list-prepend, then `Enum.reverse/1` in `audit_log/1`. O(N) reversal on each read. Fine for skeleton, but a deque or per-write index would help if audit volume grows.
- `sorted_values/1` resorts on every `list_*` call. O(N log N) per request. Fine for tens-to-hundreds; would want indexes for thousands.

Nothing blocking.

## Bottom line

Well-scoped server-side skeleton with strong defensive testing. Two structural items (namespace, binding) are easier to fix in this PR than after Phase 5 lands more modules. The clock-injection sprawl is a Phase 5+ tracking issue; the unbounded audit log is a Phase-Postgres issue. Test coverage is good for Phase 4 surface; auth + execution-endpoint guards are exemplary.

LGTM with #1 and #2 addressed.

yairfalse (commented):
## Follow-up review of `4646755` — "address namespace and bind review"

Both top concerns from the prior review are resolved, with a few nice extras on top.

### Resolved

**1. Namespace collision (was: `Sykli.Coordinator.*` overlapped the BEAM mesh role).**

Clean rename: `Sykli.Coordinator.{Application,Router,Store,Auth}` → `Sykli.TeamCoordinator.{Application,Router,Store,Auth}`. Tests, docs, and the CLI dispatch all updated. The mesh role's `Sykli.Coordinator` keeps its name. Future contributors will read `Sykli.TeamCoordinator` and immediately know which trust domain it's in. ✓

**2. Bind address mismatch (was: Bandit on `0.0.0.0`, CLI claimed `127.0.0.1`).**

Resolved cleanly:

- `@default_bind "127.0.0.1"` — localhost by default.
- `--bind ADDRESS` (and `--bind=ADDRESS`) flag for production deploys.
- `parse_bind/1` uses `:inet.parse_address/1`, which handles IPv4 and IPv6 properly.
- New `coordinator.invalid_bind` error code with a helpful hint.
- Bandit gets the resolved address via `ip: bind`.
- Startup message now reads `http://#{bind_string}:#{port}` (matches reality).
- **Bonus**: added `"warning: plain HTTP; terminate TLS at an ingress or proxy for production"` to the human startup output. Directly addresses the TLS framing concern from the security doc.
- **Bonus**: help text explicitly says `"use 0.0.0.0 only intentionally"`, and a test asserts that string is present.

### New tests

`core/test/sykli/cli/coordinator_test.exs` adds two:

- Help documents `--bind`, the `127.0.0.1` default, and the explicit-`0.0.0.0` warning.
- Invalid bind value returns a structured `coordinator.invalid_bind` JSON error.

The "asserts the warning string" test is a nice contract guard — future contributors can't silently weaken the safe-default messaging without breaking a test.

### Items I noted as non-blocking, still open (acceptable)

- `Process.sleep(:infinity)` has no SIGTERM hook. Skeleton is in-memory; in-flight requests drop on shutdown but no state is lost. Worth a `# TODO` for the Postgres slice; not required now.
- Bearer scheme is still case-sensitive. RFC 6750 says it should be insensitive; most clients use canonical capitalization, so practical impact is low.
- Audit log unbounded growth + no `GET /v1/audit-log` endpoint. Skeleton is in-memory. Add when Postgres lands.
- `with` without `else` on infallible-today GET handlers. Defensive-only nit; would prevent a future silent-hang regression.

None of these are blockers for this PR.

### Bottom line

Thorough response to the review. The bind fix is genuinely better than what I asked for — adding the TLS warning to the startup output and asserting it via test means the safe-default messaging is now part of the contract, not just a comment. The namespace rename was mechanical but exactly the right call before more modules accumulate.

LGTM from my end (GitHub blocks self-approval; leaving as comment).

PR #198, opened 2026-05-13
Add evidence requirements V1

yairfalse:
Summary:
- Adds v4 evidence_required contract support across schema, parser/validator, runtime result propagation, and all five SDKs.
- Evaluates local file evidence refs on the local shell target and records evidence_results in history, occurrences, CLI JSON, context/query/report, and MCP task maps.
- Adds missing_evidence / unsupported_evidence_requirement_for_target semantics, docs, and conformance fixtures.

Why:
Sykli needs to distinguish a command passing from the contract proving what it claimed. This PR makes evidence references a first-class enforced contract slice without uploading logs, source, prompts, or artifact bytes.

What changed:
- core: added Sykli.EvidenceRequirement, v4 version support, parser/validator checks, executor evaluation, local target file evidence checks, and evidence_results propagation.
- SDKs: Go, Rust, TypeScript, Elixir, and Python emit evidence_required and auto-detect version 4.
- tests/conformance: added 25-evidence-required plus negative schema fixtures.
- docs: documented evidence requirements, v4 versioning, and new error codes.

Compatibility:
- Existing fields remain in place; old run history without evidence_results still loads.
- No remote evidence upload and no coordinator artifact/log storage.
- No SDK behavior changes except emitting version 4 when evidence_required is declared.

Tests:
- mix format --check-formatted
- mix credo --strict
- mix test --max-cases 1
- targeted core evidence/history tests
- npm run build
- npm test -- --run
- env PYTHONPATH=src python3.14 -m pytest
- env GOCACHE=/private/tmp/sykli-go-build go test ./...
- cargo test
- sdk/elixir mix test
- env GOCACHE=/private/tmp/sykli-go-build SYKLI_CONFORMANCE_PYTHON=python3.14 tests/conformance/run.sh
- jq empty for schema and new fixtures

Note: local conformance schema validation skipped because jsonschema is not installed for python3.14; SDK conformance itself passed 150/150.

Non-goals:
- No remote evidence upload.
- No artifact store.
- No prompt/log/source capture.
- No risk/effects model.
- No agent metadata or variance model.
- No gate redesign.
- No failure-mode learning store.

Known follow-ups:
- Risk and effects model V1 should come next, after this evidence primitive is merged.


yairfalse:
## Review

Coherent feature, vocabulary-module pattern applied cleanly, schema/engine/SDKs in lockstep. Test coverage hits the happy + missing + optional paths. A few things to address before merge.

### Bugs

- **`executor.ex:826` introduces a cache-miss atom nobody pretty-prints.** Renaming `:success_criteria_requires_execution` → `:declared_checks_require_execution` orphans `Sykli.Cache.format_miss_reason/1` (`cache.ex:196`), which still pattern-matches the old name. Cache-miss messages for tasks with `success_criteria` *or* `evidence_required` now hit the `format_miss_reason(other)` catch-all and render as raw `"declared_checks_require_execution"`. Add a clause.

- **`sdk/go/sykli.go:2396-2398` is dead code.** `validateEvidenceRequirements` ranges by value then assigns `req.visibility = "local"` — discarded. Harmless today because the only public constructor (`FileEvidence`) already sets visibility, but misleading. Drop it or move the default into emission.

- **`stat_regular_file` error message is now half-wrong.** `target/local.ex:653` still says `"symlinks are not supported for success_criteria paths"`. This helper now serves evidence checks too — symlinked evidence files emit a misleading error.

### Design / parity

- **SDK parity is unequal.** Python/TS/Elixir accept the full evidence-type enum (`file`, `log`, `attestation`, `occurrence`, `metric`, `test_report`, `artifact_ref`, `custom`) via dict/object. Go has private fields with only `FileEvidence` exposed; Rust has a single-variant enum. The other seven types are unreachable from Go/Rust public APIs. README/CLAUDE.md claim full builder parity. Either restrict the schema to `file` (the only runtime-evaluated type) or add Go/Rust builders for the rest.

- **Atom names freeze old version numbers.** `parse_task_type` and `SuccessCriteria.validate` now accept `["3", "4"]`, but the error atoms remain `:task_type_requires_version_3` / `:success_criteria_requires_version_3`. Rendered messages will keep saying "requires version 3" when a v5 pipeline hits them. Atoms are a stable wire fact in `Sykli.Validate` output — cheapest moment to rename is now. Suggest `:task_type_requires_v3_or_newer` or `:task_type_requires_semantic_version`.

- **Schema/engine asymmetry on `ref_pattern`.** `evidence_requirement.ex:558` permits `ref_pattern` on every type. Schema only requires it for `file` but doesn't forbid it elsewhere. `predicate` is correctly file-only on both sides. Either restrict `ref_pattern` to file in both, or document it as reserved for future non-file evaluators.

### Smaller nits

- **`Sykli.Error.missing_evidence/3` re-filters already-filtered results.** Caller in `target/local.ex:1156` passes `failures`; `error.ex:259` calls `EvidenceRequirement.failures(results)` again. Idempotent but the contract for what gets passed in is unclear — pick one side.

- **Variable/function shadowing.** `evidence_satisfied/_missing/_unsupported` use the `name` param and `name()` function in the same struct literal (`name: name, target: name()`). Works, but reads as a typo. Rename param to `evidence_name`.

- **CLAUDE.md regression.** Previous bullet enumerated the 11 negative version fixtures by name; new bullet just points at the directory. Fine if intentional.

- **Test gaps worth filling before this becomes load-bearing:**
  - Executor fallback path for a target without `evaluate_evidence_required/4` (K8s).
  - `evidence_required` on a containerized local task (the `%{container: container}` unsupported clause).
  - Engine-level rejection of `evidence_required` on review nodes (you have schema-invalid fixtures, not engine coverage).

### Security

Path containment, symlink rejection, and cache-bypass-when-evidence-required all carry over correctly from `success_criteria`. No new TLS/secret-masking surfaces touched.

### Recommendation

Approve once the cache miss-reason atom is wired up and the dead Go line is cleaned up. Treat SDK parity and the version-named atoms as follow-ups so we don't repeat this when v5 lands.

PR #200, opened 2026-05-15
build: validate carve-outs, --timeout enforcement

yairfalse:
## Summary

- Track A: added validator regression coverage for the distinct command: null JSON shape through Sykli.Validate.validate_json/1. The missing-command implementation was already present and keeps the existing :missing_command error type.
- Track B: added deterministic Fake runtime sleep timeout modeling and an executor regression test proving local --timeout surfaces task_timeout failure_semantics without waiting for command completion.
- Removed stale timeout guidance from CLAUDE.md. The stale command-validation claim was not present in repo CLAUDE.md; the existing local Claude memory copy was updated outside the repo.

## Verification

- Track A: cd core && mix format && mix test
- Track B: cd core && mix format && mix test

## Track C status

Not included. I stopped before implementation because Phase 8 requires design choices that are not unambiguous from the current Phase 7 pattern:

- Sykli.TeamCoordinator.Router currently states it exposes no gate endpoints, while Phase 8 requires coordinator gate request/decision APIs.
- Sykli.TeamCoordinator.Store heartbeat currently returns decisions: [], while the task allows either heartbeat delivery or explicit polling.
- CLI gate commands are explicitly local-only today, so adding --team requires choosing how sessions/tokens are discovered and how remote approval maps onto local state.

Per the task instructions, I did not guess on Track C or ship a partial implementation.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -347,5 +347,4 @@ These shape how code decisions get made. They are not aspirational — the test 
 
 ## Known Issues
 
-- `--timeout` flag doesn't enforce timeouts on local target (tasks run to completion)
 - Docker-tagged tests are skipped when no Docker daemon is available; run them explicitly with `mix test.docker`.
--- a/core/lib/sykli/runtime/fake.ex
+++ b/core/lib/sykli/runtime/fake.ex
@@ -31,6 +31,14 @@ defmodule Sykli.Runtime.Fake do
 
   Calls are still recorded when scripted.
 
+  ## Timeout modeling
+
+  When no `:run` script is provided, the Fake models only bare `sleep N`
+  commands, where `N` is an integer or fractional second value. If that
+  modeled duration exceeds `opts[:timeout_ms]`, `run/4` returns
+  `{:error, :timeout}` without actually sleeping. Compound shell commands
+  and absolute sleep paths are not parsed.
+
   ## Determinism
 
   No wall clock, no global RNG. Identifiers derive from `:erlang.phash2/1`
--- a/core/lib/sykli/runtime/fake.ex
+++ b/core/lib/sykli/runtime/fake.ex
@@ -50,7 +50,7 @@ defmodule Sykli.Runtime.Fake do
   @impl true
   def run(command, image, mounts, opts) do
     record(opts, {:run, command, image, mounts, opts})
-    scripted(opts, :run, {:ok, 0, 0, ""})
+    scripted(opts, :run, default_run(command, opts))
   end
 
   @impl true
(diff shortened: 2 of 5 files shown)


yairfalse:
## Review

### Overview
Two minimal tracks land here as separate commits. Track A adds one regression test for `command: null` (the validator code already rejected it). Track B does the real work: extends `Sykli.Runtime.Fake` to deterministically model `sleep N` commands that exceed `timeout_ms`, then adds an executor test proving the local target surfaces `task_timeout` failure semantics without waiting for the command to finish. CLAUDE.md's stale `--timeout` known-issue line is removed. Track C (Phase 8 gate sync) is intentionally deferred with clear conflict notes in the PR body.

Net diff: +81 / -2 across 5 files. No production engine code changed — only the test-runtime `Fake` and tests.

### Strengths
- **Right-sized work.** No speculative refactor. Track A is 9 lines; Track B closes a real testability gap rather than papering over a non-bug.
- **Backward-compatible Fake change.** `scripted(opts, :run, default_run(command, opts))` (fake.ex:53) preserves explicit `:run` scripts — existing tests that script `:run` still win. New behavior only kicks in when nothing is scripted.
- **Regression assertions are well-chosen.** `failure_semantics_test.exs:73-102` checks both `duration_ms < 1_000` (Fake didn't actually sleep) AND `elapsed_ms < 1_000` (test wall-clock didn't either). The `sleep 30 / timeout 100` margin is generous enough not to flake on slow CI.
- **NoWallClock posture respected.** `System.monotonic_time(:millisecond)` in the executor test is allowed — CLAUDE.md scopes the check to mesh transport/fixtures, and the doc explicitly calls out that elapsed-time measurement is OK in executor tests.

### Issues and risks
- **Dogfood CI is failing.** The \`Sykli CI (Dogfood)\` check (run 25944572621) failed while \`Core Tests\` and all five SDK suites passed. Could be unrelated infrastructure, could be a real interaction with the new Fake behavior under the dogfood pipeline. **Investigate before merge** — fail-on-merge is the wrong default for a \`sykli\`-runs-itself signal.
- **\`fake_sleep_ms/1\` only recognizes bare \`sleep N\`.** \`fake.ex:106-112\` pattern-matches \`["sleep", duration]\` exactly. Compounds like \`sleep 30 && echo done\`, \`bash -c "sleep 30"\`, or \`/usr/bin/sleep 30\` all fall through to success. Fine for this PR's test, but a sharp edge for future test authors who'll assume timeout modeling Just Works. Worth a one-line moduledoc note saying only bare \`sleep N\` (integer or fractional seconds) is modeled.
- **\`semantics.details["duration_ms"] == 100\`** asserts the *timeout configuration value*, not actual elapsed runtime. That's the existing semantic of \`Sykli.Error.task_timeout/3\` so it's correct, but the field name overloading is a footgun for readers — a future maintainer will think this means the task ran for 100ms. Not a blocker; a comment on the assertion line would help.

### Minor suggestions
- \`fake_test.exs:60-63\` covers \`sleep 30 > 100ms\` (timeout) and \`sleep 0.01 < 100ms\` (success). Worth one more case: \`Fake.run("sleep 30", nil, [], [])\` (no \`timeout_ms\`) should succeed. That pins down the "no timeout configured" path and prevents accidental tightening later.
- \`default_run/2\` and helpers are pure — they could move into a \`Sykli.Runtime.Fake.Sleep\` submodule if more command modeling is anticipated, but that's premature today. Leave inlined.

### Security and performance
- No new attack surface. Test-only runtime; pure-function parsing; no shell exec.
- Float parsing on arbitrary command strings has no DoS risk — bounded by command length, no regex backtracking.

### Track C deferral
Reading the PR body, the three flagged unknowns (gate endpoints, decision delivery shape, \`--team\` token discovery) are all real design decisions, not implementation hesitations. The choice to ship A+B and stop was correct. Those decisions have since been resolved separately and are unblocked for a follow-up PR.

### Verdict
**Approve once dogfood CI is investigated.** If the failure is unrelated infrastructure (re-run green), merge. If it's a real interaction with the new Fake behavior, fix in this PR before merging.

Issue #208, opened 2026-05-22
Broaden NoWallClock guard for DET-006


yairfalse:
Blackbox DET-006 remains expected-red in Phase A. The current Credo NoWallClock configuration does not reject a planted DateTime.utc_now file under core/lib/sykli. Planned for Monster Phase B determinism guard work.

PR #211, opened 2026-05-22
fix(determinism): stabilize contract_hash, widen NoWallClock, correct DET-003

### LINKED ISSUE #207: Fix DET-003 occurrence nondeterminism
Blackbox DET-003 remains expected-red in Phase A. Two identical runs produce .sykli/occurrence.json differences after stripping declared variable fields. Planned for Monster Phase B determinism work.

yairfalse:
## Summary
**Monster Phase B** of the audit-remediation program (`docs/audit-2026-05-22.md`). Stacked on #210 (Phase A).

The investigation produced an honest course-correction: **two of the audit's three "determinism bugs" were not engine bugs.** Findings and fixes:

- **B4 — `contract_hash` stability (real bug, fixed).** `from_json/1` now recursively sorts object keys before hashing, so semantically-identical contracts hash identically regardless of map iteration order. Maps with >32 keys use a hashed representation whose iteration order is **not stable across OTP versions** — the latent hazard. Tests cover key-order independence including the >32-key case.
- **B3 — `NoWallClock` guard widened (DET-006, fixed).** The audit's "broaden to the whole engine" framing was infeasible: there are **83 legitimate wall-clock uses** across ~30 files (occurrence timestamps, run history, cache mtimes, OIDC expiry). Broadening wholesale would flag all of them. Instead the guard now covers the **pure contract & output-shaping transforms** (graph parsing/validation, the vocabulary modules, contract hashing, occurrence serializer/enrichment) — where wall-clock genuinely *is* a determinism bug. Credo now lints 32 files (was 10), still green. Time-stamping engine modules stay intentionally exempt.
- **B2 — DET-003 was a flawed test, not an engine bug.** Reproduced and proved: in two fresh workspaces the occurrence output is **byte-identical** except the run_id embedded in the `log` path. The prior case reused `.sykli/` (so run 2 was a cache hit, since the cache lives globally at `~/.sykli/cache`) and stripped run_id from labels but not from the log path. Corrected to compare two fresh workspaces and strip the run_id-bearing log path. The engine is deterministic.

DET-003 and DET-006 now pass **without** `expected_failure` flags (closes #207, #208).

## Descoped (honest)
**B1 (injectable clock/id seam)** is deferred to Phase H/DST. Its original justification was fixing DET-003; since the engine is already deterministic it fixes no shipping bug, and the seam belongs where deterministic simulation actually exercises it. Per the program's own rule: don't invent a fix for a non-problem.

## Verification
| Gate | Result |
|------|--------|
| `mix format --check-formatted` | clean |
| `mix credo` | no issues (now 32 files) |
| `mix test` | **1756 tests, 0 failures** (+2 contract_hash) |
| `mix escript.build` | builds |
| `test/blackbox/run.sh` | **166 passed, 0 failed**, GH-004 expected-red (Phase C) |
| DET-003 / DET-006 | pass without flags |


test/blackbox/dataset.json
```diff
@@ -1289,14 +1289,12 @@
...
       "fixture": "single_task",
-      "command": "sykli >/dev/null && jq 'del(.id,.timestamp,.context.labels[\"sykli.run_id\"],.data.tasks[].duration_ms,.data.duration_ms,.history.duration_ms,.history.steps[].duration_ms,.history.steps[].timestamp)' .sykli/occurrence.json > a.json && sykli >/dev/null && jq 'del(.id,.timestamp,.context.labels[\"sykli.run_id\"],.data.tasks[].duration_ms,.data.duration_ms,.history.duration_ms,.history.steps[].duration_ms,.history.steps[].timestamp)' .sykli/occurrence.json > b.json && cmp -s a.json b.json",
+      "command": "mkdir -p r1 r2 && cp sykli.exs r1/ && cp sykli.exs r2/ && (cd r1 && sykli >/dev/null) && (cd r2 && sykli >/dev/null) && jq 'del(.id,.timestamp,.context.labels[\"sykli.run_id\"],.data.tasks[].duration_ms,.data.duration_ms,.history.duration_ms,.history.steps[].duration_ms,.history.steps[].timestamp,.data.tasks[].log)' r1/.sykli/occurrence.json > a.json && jq 'del(.id,.timestamp,.context.labels[\"sykli.run_id\"],.data.tasks[].duration_ms,.data.duration_ms,.history.duration_ms,.history.steps[].duration_ms,.history.steps[].timestamp,.data.tasks[].log)' r2/.sykli/occurrence.json > b.json && cmp -s a.json b.json",
```
yairfalse: DET-003 now uses fresh workspaces, which is the right direction, but this normalization deletes `.data.tasks[].log` entirely. That makes the test blind to deterministic regressions in the log evidence path itself, even though the field is part of the occurrence shape we want downstream tools to trust. Can we normalize just the variable run-id segment instead, e.g. with `walk(if type == "string" then gsub("[0-9A-HJKMNP-TV-Z]{26}"; "<run_id>") else . end)` or a targeted replacement on `.data.tasks[].log`, and keep the `log` key/value under comparison?

PR #216, opened 2026-05-25
fix(security): SSRF-guard gate webhooks (Phase C, C4)

yairfalse:
**Monster Phase C — item C4** from `docs/audit-2026-05-22.md`.

A pipeline-declared gate `webhook_url` was POSTed with no destination validation, so it could be pointed at the cloud metadata endpoint (`169.254.169.254`) or internal services from inside the trust boundary — a classic SSRF / metadata-exfiltration sink.

## Fix
- Extracted the **existing** notification-webhook SSRF guard into a shared `Sykli.HTTP.check_ssrf/1` (resolves the host; blocks loopback, link-local incl. `169.254.0.0/16`, and RFC1918 — IPv4 and IPv6).
- Applied it in `gate_service`'s `wait_webhook` (blocked → `{:denied, "webhook_url blocked: ..."}`).
- `notification_service` now **delegates** to the same helper (DRY — both webhook paths enforce identical blocking).

## Verification
`mix format` / `mix credo` (clean) / `mix test` (**1757, 0 failures**, +5 SSRF tests) / `mix escript.build`; black-box GATE cases pass.

## Phase C remaining
- **C5** Shell-runtime trust ADR + GH-004 rewrite (decision: document Shell = trusted) — own PR (establishes `docs/adr/`)
- **C3** mask file/OIDC-resolved secrets — own PR
- **C2** per-team coordinator authz (decision: enforce per-team tokens) — own PR (auth-model change)

### DIFF
--- a/core/lib/sykli/http.ex
+++ b/core/lib/sykli/http.ex
@@ -4,6 +4,8 @@ defmodule Sykli.HTTP do
   Provides TLS verification options for HTTPS endpoints.
   """
 
+  import Bitwise
+
   @doc """
   Returns SSL options for :httpc that verify server certificates and hostnames.
   """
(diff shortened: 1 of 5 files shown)


core/lib/sykli/http.ex
```diff
@@ -63,4 +63,53 @@ defmodule Sykli.HTTP do
...
+  defp private_ip?({127, _, _, _}), do: true
+  defp private_ip?({10, _, _, _}), do: true
+  defp private_ip?({172, b, _, _}) when b >= 16 and b <= 31, do: true
+  defp private_ip?({192, 168, _, _}), do: true
+  defp private_ip?({169, 254, _, _}), do: true
+  defp private_ip?({0, 0, 0, 0}), do: true
+  defp private_ip?(_), do: false
+
+  defp private_ip6?({0, 0, 0, 0, 0, 0, 0, 1}), do: true
+  defp private_ip6?({0xFE80, _, _, _, _, _, _, _}), do: true
```
yairfalse: This IPv6 matcher only blocks the exact first hextets `fe80`, `fc00`, and `fd00`, but the blocked ranges are broader: link-local is `fe80::/10` (`fe80` through `febf`) and unique-local is `fc00::/7` (`fc00` through `fdff`). I reproduced locally that `Sykli.HTTP.check_ssrf("http://[fd12:3456::1]/hook")`, `http://[fc01::1]`, and `http://[fe90::1]` all currently return `:ok`. That leaves an IPv6 private/link-local SSRF bypass. Please compare the first 16-bit segment with masks/ranges, e.g. `(h &&& 0xffc0) == 0xfe80` for link-local and `(h &&& 0xfe00) == 0xfc00` for ULA, and add tests for non-boundary addresses like `fd12::1` and `fe90::1`.

core/lib/sykli/http.ex
```diff
@@ -63,4 +63,53 @@ defmodule Sykli.HTTP do
...
+  @spec check_ssrf(String.t()) :: :ok | {:error, String.t()}
+  def check_ssrf(url) do
+    case URI.parse(url).host do
+      nil ->
+        {:error, "URL has no host"}
+
+      host ->
+        host_charlist = String.to_charlist(host)
+
+        case :inet.getaddr(host_charlist, :inet) do
```
yairfalse: For DNS names, checking a single `:inet.getaddr/2` result is weaker than the request path: a hostname can resolve to multiple A/AAAA records, and this guard accepts the URL if the one sampled address is public even if another returned address is private/link-local. `:httpc` resolves the host again when sending, so it may connect to a different record than the one checked. For an SSRF guard this should reject if *any* resolved address is private, using `:inet.getaddrs(host, :inet)` and `:inet.getaddrs(host, :inet6)` and checking the full returned sets (ideally treating mixed public/private answers as blocked).

PR #224, opened 2026-06-11
docs(readme): show typed failure semantics — the agent demo section

yairfalse:
## What

Phase 1 (strategy) opener — the README now demonstrates the differentiator instead of only claiming it:

- New section **"When It Fails, Agents Read Facts — Not Logs"** right after the quickstart: the per-task `failure_semantics` + `agent_hints` JSON an agent sees via `sykli run --json` / `report --json` / MCP
- The JSON block is **captured from a real run** (v3 pipeline, `file_exists` criterion failing → `criteria_failure`), verified against the binary before committing
- README version strings 0.6.1 → 0.7.0, with an honest availability note: Go module live now, other registries catching up
- Contributing section links CONTRIBUTING.md

No longer gated on registry publishing — the availability note keeps the table truthful either way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,6 +316,10 @@ Requires Elixir 1.14+.
 | Elixir | `{:sykli_sdk, "~> 0.7.0"}` in `mix.exs` | `sykli.exs` |
 | Python | `pip install sykli==0.7.0` | `sykli.py` |
 
+The Go module is live today. Registry packages for the other SDKs are
+catching up to 0.7.0 — until your registry shows it, every SDK is usable
+directly from `sdk/<lang>/` in this repository at the `v0.7.0` tag.
+
 All SDKs emit the same canonical contract shape.
 
 ## Capabilities
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ func main() {
 ```
 
 ```text
-sykli · pipeline.go                                local · 0.6.1
+sykli · pipeline.go                                local · 0.7.0
 
   ●  test     go test ./...                        108ms
   ●  build    go build -o app                      612ms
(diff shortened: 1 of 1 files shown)


dimashev (commented):
**Review — content LGTM (docs-only, accurate).** Cross-checked the new failure-semantics section against the code on `main`; everything matches:

- Version `0.7.0` matches `VERSION`, and the `v0.7.0` tag exists — so the "usable from `sdk/<lang>/` at the v0.7.0 tag" note holds.
- The `failure_semantics` JSON matches `failure_semantics.ex` — `class / retryable / source / reason / message / details` (`details` is a real struct field, defaults to `%{}`).
- Listed classes (`runtime_failure`, `criteria_failure`, `timeout`, `dependency_failure`, `policy_block`, `missing_evidence`) are all real; "…and friends" fairly covers the rest.
- The example is internally consistent with the logic: `criteria_failure` → `source: "criteria"`, `retryable: false`, and its `agent_hints` (only `inspect_contract: true`) matches `AgentHints`' criteria-failure mapping exactly.
- `agent_hints`' six fields match the `@empty` map field-for-field.
- Referenced files all exist: `docs/failure-semantics.md`, `docs/agent-readable-failure-output.md`, `CONTRIBUTING.md`.

**Minor (optional, non-blocking):** the SDK install table shows pinned `0.7.0` commands (`npm install sykli@0.7.0`, `pip install sykli==0.7.0`) as if published, while the note just below says non-Go registries are still catching up to 0.7.0. A reader could copy those before the registry has them — consider marking the non-Go rows "(when published)" to match the note.

—
🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

PR #269, opened 2026-07-03
feat(cli): add contract lock

yairfalse:
## Summary

- Adds `sykli lock` and the `sykli.lock` format with the full canonical contract.
- Adds double-emit determinism checks for lock creation and lock enforcement.
- Enforces an existing `sykli.lock` before `sykli validate` and `sykli run`.
- Adds structured contract error codes and black-box coverage for DET/NEG/JSON paths.

## Design decisions honored

- Lockfile stores the full canonical contract, not only the hash.
- Lock enforcement is presence-based: no lockfile means current behavior remains unchanged.
- Double-emit compares canonical hashes via `Sykli.ContractHash`.
- Lock pins the emitted contract before graph expansion.
- No SDK, schema, runtime, team coordinator, GitHub, or workflow changes.

## Verification

```text
$ cd core && mix format --check-formatted
# exit 0
```

```text
$ cd core && mix credo --strict
Checking 33 source files ...
787 mods/funs, found no issues.
```

```text
$ cd core && mix compile --warnings-as-errors --force
Compiling 218 files (.ex)
Generated sykli app
```

```text
$ cd core && mix test --seed 0
Result: 1804 passed (1 property, 1803 tests), 13 excluded
```

```text
$ cd core && mix test
Result: 1804 passed (1 property, 1803 tests), 13 excluded
```

```text
$ cd core && mix escript.build
Generated escript sykli with MIX_ENV=dev
```

```text
$ test/blackbox/run.sh --filter=DET
Passed: 8
Failed: 0
Skipped: 168
Total: 176
```

```text
$ test/blackbox/run.sh --filter=NEG
Passed: 23
Failed: 0
Skipped: 153
Total: 176
```

```text
$ test/blackbox/run.sh --filter=JSON
Passed: 11
Failed: 0
Skipped: 165
Total: 176
```

```text
$ eval/oracle/run.sh
Passed: 76
Failed: 0
Total: 76
```

## Manual demo

```text
$ sykli lock
Locked contract sha256:b5c01969762b8f04829853700729f4c139f8aa3a2a84536cfe992328a6e48636

$ sykli validate
Valid

$ perl -0pi -e 's/echo hi/echo bye/' sykli.exs
$ sykli run
error[contract.lock_mismatch]: emitted contract hash does not match sykli.lock
help: run `sykli lock` to accept the new contract edition
```

`sykli contract --diff` is intentionally not demonstrated in this PR because that command is PR 2 in the mandate.

## Out of scope

- SDK canonicalization.
- Schema v5, `actor`, `mandate`, or audit work.
- CI gating and workflow changes.
- `sykli contract` / `sykli contract --diff` surface, which is the next PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **`sykli lock`.** Writes `sykli.lock` with the canonical contract and
+  enforces matching locked contracts in `sykli validate` and `sykli run`.
 - **Daemon heartbeat loop (Monster Phase D, PR 1).** New
   `Sykli.Daemon.Heartbeat` GenServer implements the runtime side of
   `docs/daemon-join-protocol.md`: coordinator-owned cadence
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -210,6 +210,7 @@ Local-first is binding: anything new must work in mode 1 with no network. The co
 | `Daemon.Join` / `Daemon.SessionStore` | `daemon/join.ex`, `daemon/session_store.ex` | Outbound daemon join + heartbeat protocol (see `docs/daemon-join-protocol.md`). Session token persisted under `.sykli/`. |
 | `ReviewPrimitive` | `review_primitive.ex` | Deterministic dispatch for `kind: "review"` nodes. Canonical names only (e.g. `api_breakage`); hyphenated aliases are rejected. Returns `Result{review_type, status, severity, message, tool, findings, evidence}`. Unsupported primitives fail explicitly — never silently skipped. |
 | `ContractHash` | `contract_hash.ex` | `sha256:` hashes for emitted SDK JSON. Canonicalizes by re-encoding parsed JSON so formatting and SDK-side comments don't perturb the hash. Used as Team Mode contract identity. |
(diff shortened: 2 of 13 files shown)


yairfalse:
Re-verified 89cc4a8 against a fresh build: both original repros are fixed (empty dir → `sdk_not_found` exit 1, no crash; failing emission → `sdk_failed`, no false nondeterminism; true nondeterministic fixture still correctly rejected). Good.

But **Oracle Eval is now red on case 018** ("Validation: no SDK file exits with error"), and it's a real regression, not a flaky check: wrapping the detector failure in `verify_project/2` means lock enforcement now *reports* the missing-SDK error itself, pre-empting the normal `validate`/`run` paths that own that message. The user-visible error text changed, and the oracle pinned the old text as ground truth.

Fix shape (from the original review comment): a detector `find` failure means there is *nothing to lock-enforce* — return `:ok` and fall through, so the established paths report the missing file with their existing wording:

```elixir
def verify_project(path, opts \\ []) do
  case detector(opts).find(path) do
    {:ok, {sdk_file, _runner} = sdk} ->
      # existing lockfile-present/absent logic
    {:error, _reason} ->
      :ok   # no project detected — normal validate/run paths own this error
  end
end
```

Keep the `Error.wrap` improvement inside `double_emit` — that one is correct and stays.

**Do not adjust oracle case 018 to match the new message** — the oracle is ground truth and that edit would be exactly the test-weakening move this whole feature exists to catch.

PR #270, opened 2026-07-03
feat(cli): add contract surface

yairfalse:
## Summary

- add `sykli contract` and `sykli contract --diff` with shared JSON envelopes
- add pure contract change classification for weakening/strengthening/neutral deltas
- add focused ExUnit and black-box coverage for render, JSON, and weakening diff exit behavior

## Stack note

This PR is based on PR #269 (`feature/contract-lock`), which is currently unmerged. The branch targets `main` per the brief, so the GitHub diff includes the PR #269 base work until that lands.

## Verification

- `cd core && mix format --check-formatted`
- `cd core && mix credo --strict`
- `cd core && mix compile --warnings-as-errors --force`
- `cd core && mix test test/sykli/contract_diff_test.exs`
- `cd core && mix test --seed 0`
- `cd core && mix test`
- `cd core && mix escript.build`
- `jq empty test/blackbox/dataset.json`
- `test/blackbox/run.sh --filter=DET`
- `test/blackbox/run.sh --filter=NEG`
- `test/blackbox/run.sh --filter=JSON`
- `eval/oracle/run.sh`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **`sykli lock`.** Writes `sykli.lock` with the canonical contract and
   enforces matching locked contracts in `sykli validate` and `sykli run`.
+- **`sykli contract`.** Renders the current emitted contract and supports
+  `--diff` against `sykli.lock` or another lockfile with weakening
+  classification.
 - **Daemon heartbeat loop (Monster Phase D, PR 1).** New
   `Sykli.Daemon.Heartbeat` GenServer implements the runtime side of
   `docs/daemon-join-protocol.md`: coordinator-owned cadence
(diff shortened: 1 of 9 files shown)


yairfalse:
Review against the brief: classifier implements the table faithfully with per-row tests plus the `classify(a,a) == []` fixture property; `--diff` exits 1 on weakening; guardrails B-E.2 updated to `partial` exactly as specified. Inherits the two blocking bugs commented on #269 (shared `ContractLock`/CLI plumbing) — fix there and rebase. Two minor findings of its own, neither blocking:

**1. Missing lockfile reuses `contract.lock_corrupt`.** `diff_lock_path/3` reports an *absent* `sykli.lock` with the corrupt code. The hint is right, the code is misleading. Either extend the `contract.lock_corrupt` description in `docs/error-codes.md` to explicitly cover "missing", or (cleaner) add `contract.lock_missing` to the catalog.

**2. Unclassified field changes are invisible.** Changes to fields outside the classification table (`timeout`, `image`, `outputs`, removal of a `gate` field from a task that still exists, …) emit no change entry at all — so `sykli contract --diff` can print "No contract changes" while `from_hash != to_hash`, which reads as a contradiction. Suggest a catch-all: when two versions of a task differ after the classified fields are accounted for, emit one neutral `task changed (other fields)` entry. Keeps the brief's "unlisted defaults to neutral" rule literally true.

One decision worth recording in the PR body: `--diff --json` with weakening returns `ok: true` + exit 1 (domain verdict drives the exit code, envelope reports command success). That reads as the right call — consistent with the failure-with-data philosophy — but it should be documented so the next JSON surface copies it deliberately.

